Issue #6, opened 2018-07-07
The response is never passed


Xiot:
**Describe the bug**
The 'response' object is never passed into setState, so it will never be available through the render prop.

**To Reproduce**
https://github.com/contiamo/restful-react/blob/master/src/index.tsx#L188

**Expected behavior**
The response object should either be passed into setState, or removed from the render prop.

TejasQ:
Thanks for the find! Seems like a trivial fix! Wanna submit a PR? I’d be happy to merge!

Xiot:
I tried putting it in locally and my unit tests kept freezing up. I haven't set up a react page yet to test it out for real. 
Also setState isn't 'await'able

TejasQ:
Actually I’m pretty sure [`setState` is asynchronous](https://medium.com/@wereHamster/beware-react-setstate-is-asynchronous-ce87ef1a9cf3).

TejasQ:
Right! My bad. Good catch!

TejasQ:
Fixed in #7

Issue #2, opened 2018-07-02
Add CodeSandbox examples and link in README.md


TejasQ:
# Why
Proper, real-world examples (maybe using the [Dogs API](https://dog.ceo)) would help people understand the power of this tool and how it works fully: including examples for state handling, polling, and other advanced features of this library.

# How
Head on over to [CodeSandbox](https://codesandbox.io), create examples, and submit a PR with links at relevant places in the README.

PR #7, opened 2018-07-09
Update implementation for cleaner mutations

TejasQ:
# Why

- We were not resetting error states on refetch (!)
- We decided to rename `host` to `base`
- I was incorrectly using `setState` (thanks, @Xiot!)
- `lint-staged` was incorrectly configured and did not work on precommit.
- `<Get lazy` and then calling `post` inside its body seemed like a workaround.
- Errors were being caught and handled _inside_ the library when instead how they're handled should be up to consumers.
- `index.tsx` exported a whole bunch of other things, while containing the GET definition.

## Changes
- Better error handling, now passing response (#6)
- Renamed host prop to base (#2)
- Removed useless `console.log`
- Cleaned up precommit hook (it actually works now)
- `<Mutate` component that handles HTTP verbs that are not `GET`.
- Errors are thrown and left for users to handle in a `catch` block.
- Cleaner file structure.

### This will result in a new major version because of API breakage.


src/index.tsx
```diff
@@ -86,8 +86,10 @@ export interface GetComponentProps<T> {
   /**
    * An escape hatch and an alternative to `path` when you'd like
    * to fetch from an entirely different URL..
+   *
+   * @deprecated Deprecated in favor of a `base` prop (https://github.com/contiamo/restful-react/issues/4)
```
Xiot: This deprecation warning doesn't make sense on the `base` prop.

PR #8, opened 2018-07-09
Update HTTP verbs in mutations to be uppercase

TejasQ:
# Why
Turns out request methods are case sensitive. Who knew?!


fabien0102:
nginx knew ^^

PR #11, opened 2018-07-16
Enhance/compose paths

TejasQ:
# Why

Currently, when using `Poll`, every `Get` inside it has its path appended to whatever `Poll`'s path was. _But why?_ Typically, a user would want to poll a specific endpoint and _not_ have it affect subsequent `GET` requests. This PR solves this case.

It also removes the possibility of nesting `Poll` paths, which generally is a bad practice. _Why would anyone nest Polls?_

### DIFF
--- a/src/Get.tsx
+++ b/src/Get.tsx
@@ -61,7 +61,7 @@ export interface GetComponentProps<T = {}> {
    * A function to resolve data return from the backend, most typically
    * used when the backend response needs to be adapted in some way.
    */
-  resolve?: ResolveFunction<T>;
+  resolve: ResolveFunction<T>;
   /**
    * Should we wait until we have data before rendering?
    * This is useful in cases where data is available too quickly
@@ -109,6 +109,10 @@ class ContextlessGet<T> extends React.Component<GetComponentProps<T>, Readonly<G
     return { loading: !props.lazy };
   }
 
+  public static defaultProps = {
+    resolve: (noop: any) => noop,
+  };
+
   public componentDidMount() {
     if (!this.props.lazy) {
       this.fetch();
@@ -155,11 +159,8 @@ class ContextlessGet<T> extends React.Component<GetComponentProps<T>, Readonly<G
   };
 
   public fetch = async (requestPath?: string, thisRequestOptions?: RequestInit) => {
-    const { base, path } = this.props;
+    const { base, path, resolve } = this.props;
     this.setState(() => ({ error: "", loading: true }));
-
-    const { resolve } = this.props;
-    const foolProofResolve = resolve || (noop => noop);
     const response = await fetch(`${base}${requestPath || path || ""}`, this.getRequestOptions(thisRequestOptions));
 
     if (!response.ok) {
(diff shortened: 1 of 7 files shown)


TejasQ:
I can do one better and clean up the lint fix. Just `commit --amend`ed for your reading pleasure.

src/Get.tsx
```diff
@@ -96,28 +98,37 @@ export interface GetComponentState<T> {
...
     response: null,
     error: "",
-    loading: this.shouldFetchImmediately(),
+    loading: false,
   };
 
-  componentDidMount() {
-    this.shouldFetchImmediately() && this.fetch();
+  public static getDerivedStateFromProps(props: Pick<GetComponentProps, "wait" | "lazy">) {
+    return { loading: !props.lazy };
```
fabien0102: You don't use `wait` anymore? and what is `wait` btw? ^^

src/Get.tsx
```diff
@@ -96,28 +98,37 @@ export interface GetComponentState<T> {
...
+    }
   }
 
-  componentDidUpdate(prevProps: GetComponentProps<T>) {
+  public componentDidUpdate(prevProps: GetComponentProps<T>) {
     // If the path or base prop changes, refetch!
     const { path, base } = this.props;
     if (prevProps.path !== path || prevProps.base !== base) {
-      this.shouldFetchImmediately() && this.fetch();
+      if (!this.props.lazy) {
```
fabien0102: Why not use `this.state.loading` here?

src/Get.tsx
```diff
@@ -143,12 +154,12 @@ class ContextlessGet<T> extends React.Component<GetComponentProps<T>, Readonly<G
...
   };
 
-  fetch = async (requestPath?: string, thisRequestOptions?: RequestInit) => {
+  public fetch = async (requestPath?: string, thisRequestOptions?: RequestInit) => {
     const { base, path } = this.props;
     this.setState(() => ({ error: "", loading: true }));
 
     const { resolve } = this.props;
-    const foolProofResolve = resolve || (data => data);
+    const foolProofResolve = resolve || (noop => noop);
```
fabien0102: Technically, `noop` means no operation, and it's a classic function name that represent `() => {}`.

I think that I prefer the old `data => data` here ;)

src/Get.tsx
TejasQ: I tried to be clear about what `wait` is [here](https://github.com/contiamo/restful-react/pull/11/commits/64b391ccac6bc1142b93e4aec70424dd8eaca1f6#diff-63d62a288bee27bda4e83b106d69c46dR66). 

I wonder how we can make it more clear. 🤔 

Basically, if `wait` is set, it doesn't render anything until there is data.

src/Get.tsx
TejasQ: Because the logic flow here is simply: 

- did something change? 🤔 
- does the user want to lazy load? 🤔 
- DONT DO ANYTHING.

It doesn't quite depend on loading state.

fabien0102: Sure, but if I'm in `loading`, that means that I have a request pending, so 2 (or more) requests can arrive in this case -> carefull to concurrency 💥

TejasQ: Let's tackle this in a separate PR. I made an issue (#12) where we can handle this.

PR #13, opened 2018-07-18
Refactor URL handling to account for query params

TejasQ:
# Why
Currently, when requesting `/collection/resource?filter=blah`, nested `Get`, `Mutate` and `Poll` components naïvely append things _past the query string_. This leads to incorrect behavior.

This PR makes URL handling more reliable by treating query strings as they should be treated.

## Side effects
This PR also makes the following optimizations/adjustments to the codebase: 
- Adds missing TSLint dependencies (how was the linter working? 🤔)
- Adds better handling of `defaultProps`.
- Removes unnecessary `getDerivedStateFromProps` because we only want _initial_ state from props since `Get`s are not _fully controlled_.


src/Get.tsx
```diff
@@ -102,13 +108,9 @@ class ContextlessGet<T> extends React.Component<GetComponentProps<T>, Readonly<G
...
     error: "",
-    loading: false,
+    loading: !this.props.lazy,
   };
 
-  public static getDerivedStateFromProps(props: Pick<GetComponentProps, "lazy">) {
-    return { loading: !props.lazy };
-  }
-
   public static defaultProps = {
```
fabien0102: You should type your defaultProps here

src/Get.tsx
```diff
@@ -159,9 +161,11 @@ class ContextlessGet<T> extends React.Component<GetComponentProps<T>, Readonly<G
   };
 
   public fetch = async (requestPath?: string, thisRequestOptions?: RequestInit) => {
-    const { base, path, resolve } = this.props;
+    const { base, path, resolve } = this.props as PropsWithDefaults<T>;
```
fabien0102: So bad… we can fix this together ;)

PR #19, opened 2018-08-08
Fix Delete action without id

fabien0102:
# Why

Actually, if no id is provided for a DELETE operation, we call `${base}${path}/undefined`, so we broken some implementations.

### DIFF
--- a/src/Mutate.test.tsx
+++ b/src/Mutate.test.tsx
@@ -1,0 +1,80 @@
+import "isomorphic-fetch";
+import "jest-dom/extend-expect";
+import nock from "nock";
+import React from "react";
+import { cleanup, render, wait } from "react-testing-library";
+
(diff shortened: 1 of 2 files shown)


src/Mutate.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+      await wait(() => expect(children.mock.calls.length).toBe(3));
+
+      // transition state
+      expect(children.mock.calls[1][1].loading).toEqual(true);
+
+      // after delete state
+      expect(children.mock.calls[2][1].loading).toEqual(false);
+    });
+
+    it("should call the correct url without id", async () => {
```
fabien0102: This is the failing test that I fixed in this PR, before we have `/undefined` in this case :wink:

PR #21, opened 2018-08-08
Update Docs

TejasQ:
# Why
So that people can read and understand the project 😄


TejasQ:
[Document Restful React before changes get lost](https://trello.com/c/seazk8gH/366-document-restful-react-before-changes-get-lost)

PR #27, opened 2018-08-14
4.0.1

TejasQ:
# Why
Need correct polling spec.


src/Poll.tsx
```diff
@@ -201,7 +201,7 @@ class ContextlessPoll<TData, TError> extends React.Component<
       ...requestOptions,
 
       headers: {
-        Prefer: `wait=${wait};${lastPollIndex ? `index=${lastPollIndex}` : ""}`,
+        Prefer: `wait=${wait}s;${lastPollIndex ? `index=${lastPollIndex}` : ""}`,
```
peterszerzo: Is the `wait` key inside the header among people using this library? If not, `waitFor` sounds more readable.

TejasQ: It's actually [the Contiamo spec](https://github.com/contiamo/restful-react#long-polling). 
[Internal link, non-contiamo people look away](https://www.notion.so/contiamo/Event-Streaming-Long-Polling-78c1905334314b32beff25c412a07281).

peterszerzo: Makes sense - for such a small naming difference we can leave it as it is.

PR #28, opened 2018-08-14
Open api import

fabien0102:
# Why

The idea is to be able to give any open-api specs file (in json or yaml) and generate fully typed restful-react components 🎉 

## Usage

```sh
$ npm i -g restful-react
$ restful-react import open-api.yaml --output myAPI.d.ts
```

## Output (current)
```ts
/* Generated by restful-react */

import qs from \\"qs\\";
import React from \\"react\\";
import { Get, GetProps, Mutate, MutateProps } from \\"restful-react\\";

export type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;

export type Pet = NewPet & {id: number};

export interface NewPet {name: string; tag?: string}

export interface Error {code: number; message: string}

export type FindPetsProps = Omit<GetProps<Pet[], Error>, \\"path\\"> & {tags?: string[]; limit?: number};


export const FindPets = ({tags, limit, ...props}: FindPetsProps) => (
  <Get<Pet[], Error>
    path={\`/pets?\${qs.stringify({tags, limit})}\`}
    base=\\"http://localhost\\"
    {...props}
  />
);


export type AddPetProps = Omit<MutateProps<Error, Pet, NewPet>, \\"path\\">;


export const AddPet = (props: AddPetProps) => (
  <Mutate<Error, Pet, NewPet>
    path={\`/pets\`}
    base=\\"http://localhost\\"
    {...props}
  />
);


export type FindPetByIdProps = Omit<GetProps<Pet, Error>, \\"path\\"> & {id: number};


export const FindPetById = ({id, ...props}: FindPetByIdProps) => (
  <Get<Pet, Error>
    path={\`/pets/\${id}\`}
    base=\\"http://localhost\\"
    {...props}
  />
);


export type DeletePetProps = Omit<MutateProps<Error, void, void>, \\"path\\"> & {id: number};


export const DeletePet = ({id, ...props}: DeletePetProps) => (
  <Mutate<Error, void, void>
    path={\`/pets/\${id}\`}
    base=\\"http://localhost\\"
    {...props}
  />
);
```

## Usage inside the client project (current)

```ts
import { FindPetById } from "./myAPI"

const MyComponent = () =>
  <FindPetById id={42}>
   {(data) => /* I have type definition here!!! */}
  </FindPetById>
```

## Todo

As the above example show, everything open-api definition pattern are not following yet. I will update this example during the API evolution :wink:

- [x] Deal with `allOf` pattern
- [x] Deal with parameters (`in: "path"` and `in: "query"`)
- [x] Add a Generic into `Mutation` component to be able to type `body`
- [x] Generate `Mutation` component
- [x] Add integration test
- [x] Fix the build ( we need to have `"module": "commonjs"` )
- [x] Support swagger 2 out of the box (https://github.com/Mermade/oas-kit/blob/master/packages/swagger2openapi/README.md)
- [x] The `verb` is missing in `Mutate`
- [x] Add quick documentation
- [x] Try in real project to see if the typescript definition follow as expected


fabien0102:
[OpenAPI import / generator](https://trello.com/c/hrzdU7RL/407-openapi-import-generator)

bin/restful-react-import.js
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env node
+
+const program = require("commander");
+const { join } = require("path");
+const importOpenApi = require("../dist/import-open-api");
```
TejasQ: 🤔

scripts/import-open-api.ts
```diff
@@ -0,0 +1,236 @@
...
+ */
+const generateGetComponent = (operation: OperationObject, verb: string, route: string) => {
+  if (!operation.operationId) {
+    throw new Error(`Every path must have a operationId - No operationId set for ${verb} ${route}`);
+  }
+
+  const componentName = pascal(operation.operationId!);
+
+  const isOk = ([statusCode]: [string, ResponseObject | ReferenceObject]) => statusCode.toString().startsWith("2");
+  const isError = ([statusCode]: [string, ResponseObject | ReferenceObject]) => !statusCode.toString().startsWith("2");
```
TejasQ: Careful, `.startsWith("3")` is not an error.

scripts/import-open-api.ts
```diff
@@ -0,0 +1,236 @@
...
+          return `export type ${pascal(name)} = ${getScalar(schema)}`;
+      }
+    })
+    .join("\n\n");
+
+const main = (path: string) => {
+  const schema = importSpecs(path);
+
+  if (!schema.openapi.startsWith("3.0")) {
+    throw new Error("This tools can only parse open-api 3.0.x specifications");
```
TejasQ: We _should_ support older versions too.

bin/restful-react-import.js
fabien0102: This file is in `bin` and a part of the restful-react build, so he take the generated `.js` file from `dist` ;)

scripts/import-open-api.ts
fabien0102: Since a lot of tools can make the swagger -> open-api migration, it's not really a problem :wink:

bin/restful-react-import.js
TejasQ: This is quite confusing. Can we think of a clearer way to express this in code?

scripts/import-open-api.ts
```diff
@@ -0,0 +1,236 @@
...
+  const { ext } = parse(path);
+  return ext === ".yaml" ? YAML.parse(data) : JSON.parse(data);
+};
+
+/**
+ * Generate a restful-react Get compoment from openapi operation specs
+ *
+ * @param operation
+ */
+const generateGetComponent = (operation: OperationObject, verb: string, route: string) => {
```
restrry: file is called `import-open-api` and `generateGetComponent` consumes it. so consumer shouldn't be declared here. am I wrong?

fabien0102:
@restrry Thanks for your feedbacks 💯 This PR is totally work in progress, and sadly I don't have any time to work on it… 😞 

Ideally I will also split this huge file in smaller one, and add some unit tests for every small parts (when it make sense).

One of the big part missing where you can maybe help, is about the type of the body. Let me add a bit more context:
```js
<Mutation verb="POST">{
  (post) => <Button onClick={() => post({/* the body can't be typesafe for now */})}/>
}</Mutation>
```

To deal with this I would like to add more generic into Mutation to be able to specify the type of the body. The goal of course is to have this type generated from open-api specs (https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#request-body-object)

scripts/import-open-api.ts
```diff
@@ -0,0 +1,236 @@
...
+};
+
+/**
+ * Extract responses types from open-api specs
+ *
+ * @todo remove potential duplicates
+ * @param responses reponses object from open-api specs
+ * @param componentName name of the current component
+ */
+const getResponseTypes = (responses: Array<[string, ResponseObject | ReferenceObject]>, componentName: string) =>
```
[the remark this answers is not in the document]
fabien0102: My old me probably have a plan about this, but now… 😄 Good catch!

scripts/import-open-api.ts
fabien0102: The entry point is the `main` function on bottom, and yes I will split this file in smaller, and also add a lot of unit tests! It was just more conveniant like this to iterate :wink: (it's a wip branch)

TejasQ:
@restrry to echo what @fabien0102 said, your contributions here would be _super_ helpful – unfortunately we're quite busy developing the product so we need all the help we can get on this project. 😉

TejasQ:
@fabien0102 has all the info. He'll respond today.

fabien0102:
@restrry Logically everything in the PR description (https://github.com/contiamo/restful-react/pull/28#issue-208271696), the idea is to have the best as possible type safe component generated from open-api specs. My current implementation only have the `Get` component generated and also every types needed for this one. 

I really trying to keep the PR todo list and examples up to date (I also add this query params on the todolist).

To be very transparent, the trello card attached:
image

Really, all relevant information are in the description, the trello card is just for a planning purpose :wink:

micha-f:
@fabien0102 This is incredibly cool.

How about adding optional frontend validation of the responses in the frontend (maybe using https://www.npmjs.com/package/yup) as a second step? We could set it up so that you can enable/disable this validation and choose whether validation errors should throw or just console warn. These could be options to the generator script so that you only adding this overhead to your code if you really want to. 

@peterszerzo You have experience with `yup` - what are your thoughts?

peterszerzo:
I've had a very good time working with `yup` so far, it supports pretty much everything with a reasonably friendly API. Not quite https://package.elm-lang.org/packages/elm/json/latest/Json-Decode, but it'll serve us really well :). Making it easy to generate validations from Open API would be a huge safety win for applications, and it is really not that difficult.

(loosely related, I've proposed some `restul-react` changes that allow the library to work better with yup: https://github.com/contiamo/restful-react/issues/57)

fabien0102:
@micha-f Very good idea for adding yup! It can be a very nice second step for this generator 💯 We already have talked about this validation layer with @peterszerzo :wink:

But first of all, I need to fix the build and test everything in a real project to be sure everything is well following.

restrry:
I tried to approach the problem in a bit different way. in the chain `raw spec --> transformed data --> component declaration` the most error-prone layer is a transformation from swagger spec to js objects, so I started looking at off-the-shelf solutions.
https://github.com/swagger-api/swagger-codegen/  - probably the best transformer, but requires Java, could be problematic to maintain consistency between the peer's environment, CI etc.
https://github.com/wcandillon/swagger-js-codegen - ready solution in JavaScript, exists for 4 years, so most likely covers a majority of use cases, not actively supported. trying to figure out the current status. meantime started using the fork https://github.com/mtennoe/swagger-typescript-codegen and implemented Proof of Concept with it (for my work). there is something to improve, so already started contributing in the lib. hope we can use another template engine, but now it supports the only mustache. anyway there is an example just for demo purposes for your use case https://github.com/contiamo/restful-react/compare/master...restrry:swagger-gen?expand=1

TejasQ:
@restrry thank you for your contribution!

We actually looked into off the shelf solutions as well and, unfortunately could not find much that we could comfortable work with. I think it might be easier to process JSON swagger docs (both, v2 and OpenAPI) using Babel, which we have some experience with for the transformation layer (this is what Babel is _for_). However, this doesn't quite apply for `yaml` files and so is possibly not an ideal solution unless we `yaml -> json -> babel -> components`, which could be either cool or convoluted depending on perspective. Let's see how things develop.

> anyway there is an example just for demo purposes for your use case https://github.com/contiamo/restful-react/compare/master...restrry:swagger-gen?expand=1

This link doesn't quite work as expected. Do you have an alternative? 😅

restrry:
updated the link.
regarding babel-plugin - suspect it will be problematic, because babel supposed to be used for code transformation, not data transformation/mapping

TejasQ:
We seem to be mixing up concepts a bit, my apologies. 😅

I was talking more about the _static_ domain of transforming `openapi.json` to TypeScript interfaces (which is what [the tools you've mentioned here](https://github.com/contiamo/restful-react/pull/28#issuecomment-429083499) do, no?), and not operating on actual _data_ returned from a real-world request. I think you're right – babel would not be an ideal tool for the latter case. 

Also, I'm not sure how moustache (or any templating at all) fits into the equation. Could you help me out a bit there? I think I'm missing something...

micha-f:
@fabien0102 Might make sense to check out the `swagger-typescript-codegen` approach. It seems to avoid implementing all the possible edge cases of open api specs one by one as people run into problems.

restrry:
@TejasQ 
data transformation pipeline
`raw swagger spec ---with data parser into---> transformed data  --- with template engine into---> component declaration`
right now `swagger-typescript-codegen` doesn't allow to separate concerns and provide only one method that transforms spec in component declaration, so you have to use built-in template engine (mustache). shouldn't be hard to fix tho

fabien0102:
Every solutions can work in this case, I must admit that I don't really see the advantage to use mustache or handlebar since we have built-in template string in JavaScript, but this also works.

Just to compare:
```
{{#tsType}}
{{! must use different delimiters to avoid ambiguities when delimiters directly follow a literal brace {. }}
{{=<% %>=}}
<%#isRef%><%target%><%/isRef%><%!
%><%#isAtomic%><%&tsType%><%/isAtomic%><%!
%><%#isObject%>{<%#properties%>
'<%name%>'<%^isRequired%>?<%/isRequired%>: <%>type%><%/properties%>
}<%/isObject%><%!
%><%#isArray%>Array<<%#elementType%><%>type%><%/elementType%>>|<%#elementType%><%>type%><%/elementType%><%/isArray%><%!
%><%#isDictionary%>{[key: string]: <%#elementType%><%>type%><%/elementType%>}<%/isDictionary%>
<%={{ }}=%>
{{/tsType}}
```
https://github.com/mtennoe/swagger-typescript-codegen/blob/bfcd0cae2f8ced0dcfafc2153c40f1130cba2fd2/templates/type.mustache#L1-L12

https://github.com/contiamo/restful-react/blob/019a14732120ad3828decc4759bc100715bf05d5/src/scripts/import-open-api.ts#L281-L292


The main difference between theses libraries and my implementation is that nobody use yet OpenAPI 3.x and the generation code is not it typescript. I personally really prefer use typescript for this kind of task (autocompletion on OpenAPI specs thanks to `openapi3-ts`), so very nice to play with 😃 

Regarding the edge cases, I'm quite confident, I had the OpenAPI specs opened everytime when I develop this and a lot of unit tests. But I will discover the remaining one during my integration tests with our products 😃 

But we can borrow some ideas from these generators:
- ignore `deprecated` endpoint (to be discuss)
- use `swaggerType.$ref.substring(swaggerType.$ref.lastIndexOf('/') + 1);` to resolve $ref (for now we force to have every refs inside `#/components/schemas/` as is recommended is the specs https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#reference-object) (we don't deal with external files for now, as everybody ^^)

Bref, this implementation is working well for now. We have types on every layer and it's quite cool 😎 The only missing piece is the build step, for now you need to update the `tsconfig.json` with `"module": "commonjs",` to be able to have a working version.

@restrry If you can try locally this script with your own projects to have a feedback, this can be very helpful.

Procedure to test:
- clone the branch
- `yarn`
- Update the `module` key in `tsconfig.json` -> `"module": "commonjs",`
- `yarn build`
- `./lib/bin/restful-react.js import {openapi.yaml} -o {declaration.d.ts}`

For now it's OpenAPI 3 only, so you will probably need to convert your specs (I will try to put the convertor directly inside the script if it's possible)

restrry:
@fabien0102 unfortunately, cannot help with testing, as I don't use restful-react :) just sharing my thoughts about implementation, as I'm working on the same problem for my project (on my free time), so most likely I continue with 3rd part library (swagger-typescript-codegen, for example) so other people could reuse it for their needs. 
also, I'd prefer to use swagger@v2.0 because adding another step for transformation into open-api@v3 is just another place for bugs ;)
anyway, I continue to track the progress here, probably can borrow ideas or help with something. cheers 🍺

restrry:
@fabien0102 btw a couple weeks ago I updated `swagger-typescript-codegen` and re-wrote templates in my work projects to plain js
how it looks like for your case https://github.com/contiamo/restful-react/compare/master...restrry:swagger-gen?expand=1

fabien0102:
We also need to deal with more advanced `$ref` like this 
image

The current implementation can only deals with `$ref` inside `#/components/schemas`. Of course, to deal with more types, we need to have a good prefix strategy to avoid names collision ^^

fabien0102:
Integration test in progress! If someone want also play with this beast -> `restful-react@6.2.0-openapi-1`, have fun 😃

README.md
```diff
@@ -1,5 +1,6 @@
 # RESTful React
- [![Greenkeeper badge](https://badges.greenkeeper.io/contiamo/restful-react.svg)](https://greenkeeper.io/)
+
+[![Greenkeeper badge](https://badges.greenkeeper.io/contiamo/restful-react.svg)](https://greenkeeper.io/)
```
TejasQ:
```suggestion
```

README.md
```diff
@@ -1,5 +1,6 @@
 # RESTful React
- [![Greenkeeper badge](https://badges.greenkeeper.io/contiamo/restful-react.svg)](https://greenkeeper.io/)
+
```
TejasQ:
```suggestion
```

rollup.config.js
```diff
@@ -0,0 +1,15 @@
+const { readdirSync } = require("fs");
+
+/**
+ * Rollup configuration to build correctly our scripts (nodejs scripts need to be cjs)
+ */
+module.exports = readdirSync("lib/bin")
```
TejasQ: Is rollup really needed for this? TypeScript supports a commonjs export...

src/Mutate.tsx
```diff
@@ -15,7 +15,7 @@ export interface States<TData, TError> {
   error?: GetState<TData, TError>["error"];
 }
 
-export type MutateMethod<TData> = (data?: string | {}) => Promise<TData>;
+export type MutateMethod<TData, TBody> = (data?: string | TBody) => Promise<TData>;
```
TejasQ: `TData` represents a response body so I find `TBody` a little confusing in this case. Can we name it `TRequestBody` instead?

src/scripts/import-open-api.ts
```diff
@@ -0,0 +1,417 @@
...
+  OperationObject,
+  ParameterObject,
+  PathItemObject,
+  ReferenceObject,
+  RequestBodyObject,
+  ResponseObject,
+  SchemaObject,
+} from "openapi3-ts";
+
+// @ts-ignore - no type definition here
```
TejasQ: Can we polyfill with a `.d.ts` file and `declare module "swagger2openapi"`?

src/scripts/import-open-api.ts
```diff
@@ -0,0 +1,417 @@
...
+        }
+      });
+    } else {
+      resolve(schema);
+    }
+  });
+};
+
+/**
+ * Generate a restful-react compoment from openapi operation specs
```
TejasQ:
```suggestion
 * Generate a restful-react component from openapi operation specs
```

rollup.config.js
fabien0102: Yes, because the tsconfig is managed by operational-scripts

src/scripts/import-open-api.ts
fabien0102: I tried… It was just not working, but I will give another try

src/scripts/import-open-api.ts
```diff
@@ -267,7 +272,7 @@ export const generateRestfulComponent = (
   const needAResponseComponent = responseTypes.includes("{");
 
   // We ignore last param of DELETE action, the last params should be the `id` and it's given after in restful-react
```
micha-f: What does this mean `the last params should be the `id` and it's given after in restful-react`?

fabien0102:
[go2dts -> openAPI](https://trello.com/c/6tDbH51I/597-go2dts-openapi)

fabien0102:
[go2dts -> openAPI](https://trello.com/c/6tDbH51I/597-go2dts-openapi)

package.json
```diff
@@ -55,21 +70,32 @@
     "jest-dom": "^1.12.0",
     "lint-staged": "^7.2.0",
     "nock": "^10.0.0",
+    "openapi3-ts": "^0.12.0",
     "parcel": "^1.10.3",
     "prettier": "^1.13.5",
     "react-dom": "^16.4.2",
     "react-testing-library": "^5.0.0",
+    "request": "^2.88.0",
```
TejasQ: Is there a reason you chose this over something like [isomorphic-unfetch](https://www.npmjs.com/package/isomorphic-unfetch) or [node-fetch](https://www.npmjs.com/package/node-fetch)?

package.json
```diff
@@ -55,21 +70,32 @@
...
     "tslint-plugin-blank-line": "^0.0.9",
     "typescript": "^3.0.3"
   },
   "dependencies": {
-    "lodash": "^4.17.11",
+    "case": "^1.5.5",
+    "commander": "^2.17.1",
+    "inquirer": "^6.2.0",
+    "lodash": "^4.17.10",
+    "openapi3-ts": "^0.12.0",
```
TejasQ: This is both in `devDependencies` _and_ `dependencies`. 🤔

src/Get.tsx
```diff
@@ -297,7 +307,9 @@ class ContextlessGet<TData, TError> extends React.Component<
  * in order to provide new `parentPath` props that contain
  * a segment of the path, creating composable URLs.
  */
-function Get<TData = any, TError = any>(props: GetProps<TData, TError>) {
+function Get<TData = any, TError = any, TQueryParams = { [key: string]: any }>(
```
TejasQ: Can we use `Record<>` instead of `{ [key: string]: any }` in this case?

src/Mutate.test.tsx
```diff
@@ -358,6 +393,35 @@ describe("Mutate", () => {
 
       expect(onError.mock.calls.length).toEqual(0);
     });
+
+    it("should have the correct type definition", async () => {
```
TejasQ: I don't understand this test case. There is no assertion. 🤔

src/bin/restful-react-import.ts
```diff
@@ -0,0 +1,93 @@
+import program from "commander";
+import { existsSync, readFileSync, writeFileSync } from "fs";
+import inquirer from "inquirer";
+import { join, parse } from "path";
+import request from "request";
+import importOpenApi from "../scripts/import-open-api";
```
TejasQ:
```suggestion

import importOpenApi from "../scripts/import-open-api";
```

src/bin/restful-react-import.ts
```diff
@@ -0,0 +1,93 @@
...
+
+program.option("-o, --output [value]", "output file destination");
+program.option("-f, --file [value]", "input file (yaml or json openapi specs)");
+program.option("-g, --github [value]", "github path (format: `owner:repo:branch:path`)");
+program.parse(process.argv);
+
+(async () => {
+  if (program.file) {
+    const data = readFileSync(join(process.cwd(), program.file), "utf-8");
```
TejasQ: What happens if this fails?

src/Mutate.test.tsx
fabien0102: It's just to have a typescript error, jest fail if something is red ^^

fabien0102: But I agree, I would like to have a better pattern to try this

src/Get.tsx
fabien0102: `Record` of what? `Record` is for describe a record of an object, here it's a default value that is a dictionnary, so `{[key: string]: any}`

```ts
/**
 * Construct a type with a set of properties K of type T
 */
type Record<K extends keyof any, T> = {
    [P in K]: T;
};
```

src/scripts/import-open-api.ts
TejasQ: There's a _lot_ of English changes to come to this PR. My plan is to add a commit with updated docs/comments/etc.

IMO we can resolve this for now and then add docs at the end as a final touch.

micha-f: @TejasQ As long as we don't forget. This is not docs, this is a code comment. In my opinion, comments should be fixed with the code they comment on.

src/bin/restful-react-import.ts
fabien0102: image

This, to have a more explicit response ;)

src/scripts/import-open-api.ts
TejasQ: > My plan is to add a commit with updated docs/**comments**/etc.

We won't forget. 😄

TejasQ:
Let's merge this after we confirm the prerelease works with IdP.

PR #33, opened 2018-08-30
Allow nested relative requests

ericadamski:
and the ability to escape relative paths using absolute paths

# Why

Currently the path composition my be misaligned with user expectation by requiring the leading `/` indicating an absolute path when in reality this is required to compose relative paths. This PR normalizes all relative nested urls as well as the ability to escape absolute urls regardless of level of nesting. 

# Proposal:

We make an addition onto the `RestfulReactProviderProps` context of `url` that allows `base` to remain the same regardless of nesting (`url` will now carry the composed paths).

```typescript
export interface RestfulReactProviderProps<T = any> {
  /** The backend URL where the RESTful resources live. */
  base: string;
  /** The current path the request will attempt to resolve. */
  url?: string;
  /**
   * A function to resolve data return from the backend, most typically
   * used when the backend response needs to be adapted in some way.
   */
  resolve?: ResolveFunction<T>;
  /**
   * Options passed to the fetch request.
   */
  requestOptions?: (() => Partial<RequestInit>) | Partial<RequestInit>;
}
```

We would then make the change to the `Get` Component which allows us to carry the relative paths using the `url` prop.

```jsx
contextProps => (
   <RestfulReactProvider {...contextProps} url={`${contextProps.url || ""}/${props.path}`}>
      <ContextlessGet {...contextProps} {...props} />
   </RestfulReactProvider>
)
```

Then internally in the class we can resolve the url trailing and leading `/` issues with some utility functions and finally combine the sanitized url with the original base.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ import React from "react";
 import Get from "restful-react";
 
 const MyComponent = () => (
-  <Get path="/breeds/image/random">
+  <Get path="breeds/image/random">
     {randomDogImage => <img alt="Here's a good boye!" src={randomDogImage.message} />}
   </Get>
 );
@@ -144,7 +144,7 @@ import Get from "restful-react";
 
 export default () => (
   {/* Use the lazy prop to not send a request */}
-  <Get path="/breeds" lazy>
+  <Get path="breeds" lazy>
     {data => {
       return (
         <div>
@@ -156,7 +156,7 @@ export default () => (
 
           <h1>Random Breed</h1>
           {/* Escapes the relative path with parent: sends request to /random */}
-          <Get path="/random">
+          <Get absolute path="random">
             {image => <img alt="Random Image" src={image && image.message} />}
           </Get>
 
@@ -176,6 +176,8 @@ export default () => (
 
 From the above example, _not only_ does the path accumulate based on the nesting of each `Get`, but each `Get` _can_ override its parent with other props as well: including having _specific_ `requestOptions` if there was a valid use case.
 
+To escape form nested path accumulation you can sepcify the `absolute` prop which will ignore all parent relative paths.
+
 To opt-out of this behavior `Get` components can use an alternative URL as their `base` prop.
 
 #### [`Get` Component API](src/Get.tsx#L50-L87)
@@ -188,7 +190,7 @@ To opt-out of this behavior `Get` components can use an alternative URL as their
 
 ```jsx
 const MyAnimalsList = props => (
-  <Get path={`/${props.animal}`}>
+  <Get path={props.animal}>
     {(animals, { loading, error }) =>
       loading ? (
         <Spinner />
@@ -214,7 +216,7 @@ Within [Operational UI](https://github.com/contiamo/operational-ui), all of our 
 
 ```jsx
 const MyAnimalsList = props => (
-  <Get path={`/${props.animal}`}>
+  <Get path={props.animal}>
     {(animals, { loading, error }) =>
       loading ? (
         <Progress error={error} />
@@ -237,7 +239,7 @@ It is possible to render a `Get` component and defer the fetch to a later stage.
 [![Edit Restful React demos](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/30n66z45mq)
 
 ```jsx
-<Get path="/unicorns" lazy>
+<Get path="unicorns" lazy>
   {(unicorns, states, { get }) => (
     <div>
       <h1>Are you ready?</h1>
@@ -263,7 +265,7 @@ At the `RestfulProvider` level, _or_ on the `Get` level, a `resolve` prop will t
 ```jsx
 const myNestedData = props => (
   <Get
-    path="/this-should-be-simpler"
+    path="this-should-be-simpler"
     resolve={response => response.data.what.omg.how.map(singleThing => singleThing.name)}
   >
     {data => (
@@ -291,7 +293,7 @@ Restful React exposes an additional component called `Mutate`. These components 
 ```jsx
 const Movies = ({ dispatch }) => (
   <ul>
-    <Get path="/movies">
+    <Get path="movies">
       {(movies, states, actions) =>
         movies.map(movie => (
           <li>
@@ -327,7 +329,7 @@ RESTful React also exports a `Poll` component that will poll a backend endpoint 
 ```jsx
 import { Poll } from "restful-react"
 
-<Poll path="/deployLogs" resolve={data => data && data.data}>
+<Poll path="deployLogs" resolve={data => data && data.data}>
   {(deployLogs: DeployLog[], { loading }) =>
     loading ? (
       <PageSpinner />
@@ -357,7 +359,7 @@ In addition to the `Get` component API, `Poll` also supports:
 Below is a more convoluted example that employs nearly the full power of the `Poll` component.
 
 ```jsx
-<Poll path="/status" until={(_, response) => response && response.ok} interval={0} lazy>
+<Poll path="status" until={(_, response) => response && response.ok} interval={0} lazy>
   {(_, { loading, error, finished, polling }, { start }) => {
     return loading ? (
       <Progress error={error} />
--- a/README.md
+++ b/README.md
@@ -109,8 +109,6 @@ Here's a full overview of the API available through the `RestfulProvider`, along
 interface RestfulProviderProps<T> {
   /** The backend URL where the RESTful resources live. */
   base: string;
-  /** The current path the request will attempt to resolve. */
-  url?: string;
   /**
    * A function to resolve data return from the backend, most typically
    * used when the backend response needs to be adapted in some way.
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ export default () => (
 
 From the above example, _not only_ does the path accumulate based on the nesting of each `Get`, but each `Get` _can_ override its parent with other props as well: including having _specific_ `requestOptions` if there was a valid use case.
 
-To escape form nested path accumulation you can sepcify the `absolute` prop which will ignore all parent relative paths.
+To escape from nested path accumulation you can sepcify the `absolute` prop which will ignore all parent relative paths.
 
 To opt-out of this behavior `Get` components can use an alternative URL as their `base` prop.
 
--- a/README.md
+++ b/README.md
@@ -109,6 +109,8 @@ Here's a full overview of the API available through the `RestfulProvider`, along
 interface RestfulProviderProps<T> {
   /** The backend URL where the RESTful resources live. */
   base: string;
+  /** The current path the request will attempt to resolve. */
+  url?: string;
   /**
    * A function to resolve data return from the backend, most typically
    * used when the backend response needs to be adapted in some way.
@@ -124,6 +126,7 @@ interface RestfulProviderProps<T> {
 // Usage
 <RestfulProvider
   base="String!"
+  url="String"
   resolve={data => data}
   requestOptions={authToken => ({ headers: { Authorization: authToken } })}
 />;
@@ -150,13 +153,19 @@ export default () => (
         <div>
           <h1>Random Image</h1>
           {/* Composes path with parent: sends request to /breeds/image/random */}
-          <Get path="/image/random">
+          <Get path="image/random">
+            {image => <img alt="Random Image" src={image && image.message} />}
+          </Get>
+
+          <h1>Random Breed</h1>
+          {/* Escapes the relative path with parent: sends request to /random */}
+          <Get path="/random">
             {image => <img alt="Random Image" src={image && image.message} />}
           </Get>
 
           <h1>All Breeds</h1>
           {/* Composes path with parent: sends request to /breeds/list */}
-          <Get path="/list">
+          <Get path="list">
             {list => (
               <ul>{list && list.message.map(dogName => <li>{dogName}</li>)}</ul>
             )}
--- a/README.md
+++ b/README.md
@@ -124,7 +124,6 @@ interface RestfulProviderProps<T> {
 // Usage
 <RestfulProvider
   base="String!"
-  url="String"
   resolve={data => data}
   requestOptions={authToken => ({ headers: { Authorization: authToken } })}
 />;
--- a/src/Context.tsx
+++ b/src/Context.tsx
@@ -4,6 +4,8 @@ import { ResolveFunction } from "./Get";
 export interface RestfulReactProviderProps<T = any> {
   /** The backend URL where the RESTful resources live. */
   base: string;
+  /** The current path the request will attempt to resolve. */
+  url?: string;
   /**
    * A function to resolve data return from the backend, most typically
    * used when the backend response needs to be adapted in some way.
@@ -17,6 +19,7 @@ export interface RestfulReactProviderProps<T = any> {
 
 const { Provider, Consumer: RestfulReactConsumer } = React.createContext<RestfulReactProviderProps>({
   base: "",
+  url: "",
   resolve: (data: any) => data,
   requestOptions: {},
 });
--- a/src/Context.tsx
+++ b/src/Context.tsx
@@ -4,8 +4,6 @@ import { ResolveFunction } from "./Get";
 export interface RestfulReactProviderProps<T = any> {
   /** The backend URL where the RESTful resources live. */
   base: string;
-  /** The original backend URL where the RESTful resources live. */
-  originalBase?: string;
   /**
    * A function to resolve data return from the backend, most typically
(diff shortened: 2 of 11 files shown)


src/Get.test.tsx
```diff
@@ -101,7 +101,7 @@ describe("Get", () => {
     });
   });
 
-  describe("composing nested urls", () => {
+  fdescribe("composing nested urls", () => {
```
fabien0102: Careful! Please don't skip all the tests 😉

ericadamski:
The CI failure is related to unauthorized requests, not sure how to go about fixing this.

src/Context.tsx
```diff
@@ -4,6 +4,8 @@ import { ResolveFunction } from "./Get";
 export interface RestfulReactProviderProps<T = any> {
   /** The backend URL where the RESTful resources live. */
   base: string;
+  /** The current path the request will attempt to resolve. */
+  url?: string;
```
fabien0102: 🤔 I don't understand what it's suppose to do.

src/Get.test.tsx
```diff
@@ -102,6 +101,54 @@ describe("Get", () => {
...
+      const firstGetChildren = jest.fn(() => <Get path="/boom">{nestedGetChildren}</Get>);
+
+      render(
+        <RestfulProvider base="https://my-awesome-api.fake">
+          <Get path="plop">{firstGetChildren}</Get>
+        </RestfulProvider>,
+      );
+
+      await wait(() => {
+        expect(firstGetChildren.mock.calls).toHaveLength(2);
```
fabien0102: Don't need to wait for 2 if you wait for 3 :wink:

src/Get.test.tsx
ericadamski: 😱 Thank you for catching that 🙈

src/Get.test.tsx
```diff
@@ -102,6 +101,54 @@ describe("Get", () => {
...
+    it("should not compose if a absolute path is given", async () => {
+      nock("https://my-awesome-api.fake")
+        .get("/plop")
+        .reply(200);
+      nock("https://my-awesome-api.fake")
+        .get("/boom")
+        .reply(200);
+
+      const nestedGetChildren = jest.fn(() => <div />);
+      const firstGetChildren = jest.fn(() => <Get path="/boom">{nestedGetChildren}</Get>);
```
fabien0102: You should use `render()` as every other tests, you can directly neast components in this `render` function.

`nock` throw if the endpoint is not called (I invite you to try and see by yourself :wink:) so you can use this behaviour as unit test (if the url is not called, the test failed)

src/Context.tsx
ericadamski: Previously we were using the base to collect up the nested routes, for example if we had a base of `https://example.com` and a structure like so:

```jsx
<Get path="/first">
  {() => (<Get path="/second">{children}</Get>)}
</Get>
```

The base when rendering the nested `Get` would be `https://example.com/first` which means we lose the context of the original base url. But in order to maintain a nested url structure we need to carry the routes as we render more nested `Get`s. The url is what that is for. So instead of mutating the original base we append to a string of the relative location.

For the example above the base would remain the same `https://example.com` and our url variable will have a value of `/first`. if we decided to render a third `Get` our url will look like `/first/second` and so on.

Does that help? Is my comment in the code what is causing the confusion or is it that the url variable may not be necessary?

fabien0102: But we are on `RestfulReactProvider` so if I'm writting this:
```jsx
<RestfulReactProvider base="https://example.com">
<Get path="/plop" /> /* fetch https://example.com/plop */
</RestfulReactProvider>
```

But this:
```jsx
<RestfulReactProvider base="https://example.com" url="/plop" >
<Get path="/plop" /> /* fetch ??? */
</RestfulReactProvider>
```

On my opinion `base` and `url` are the same here, so it's really confusing

src/Get.test.tsx
ericadamski: Is it something like this that you are suggesting?

```typescript
     const parentRoute = nock("https://my-awesome-api.fake")
        .get("/plop")
        .reply(200);
      const nestedRoute = nock("https://my-awesome-api.fake")
        .get("/plop/boom")
        .reply(200);

      render(
        <RestfulProvider base="https://my-awesome-api.fake">
          <Get path="plop">{() => <Get path="boom">{() => <div />}</Get>}</Get>
        </RestfulProvider>,
      );

      await wait(() => {
        parentRoute.done();
        nestedRoute.done();
      });
```

src/Context.tsx
fabien0102: The goal in my memory was to have this:

```jsx
<RestfulReactProvider base="https://example.com">
   <Get path="/plop" /> /* fetch https://example.com/plop */
   <Get path="/plop">{() => (
      <Get path="/foo" /> /* fetch https://example.com/plop/foo */
      <Get path="foo" /> /* fetch https://example.com/foo */
    )}</Get>
</RestfulReactProvider>
```

So I don't understand why we add a `url` props to the provider.

ericadamski: I see what you are saying with the url property. I do agree that you will have unexpected behavior if you manually pass it as a prop.

My reasoning for using it was that we do not know what the original base url is once nested because it gets mutated.

so if I started off with a base url of: `https://example.com/v1/api` then have you structure:

```jsx
<RestfulReactProvider base="https://example.com/v1/api">
   <Get path="/plop" /> /* fetch https://example.com/v1/api/plop */
   <Get path="/plop">{() => (
      /* this one here is the issue, we don't know what the original base was because it gets mutated with the nested path */
      <Get path="/foo" /> /* expected fetch https://example.com/v1/api/foo  reality fetch https://example.com/v1/api/plop/foo */
      <Get path="foo" /> /* expected fetch https://example.com/v1/api/plop/foo */
    )}</Get>
</RestfulReactProvider>
```

ericadamski: the `url` prop is really internal only and shouldn't be exposed as part of the API, I can see through your example how it could cause confusion.

ericadamski: I don't want to make the assumption about what a base url looks like, ie.

`https://example.com/v1/api` is a valid base.

Is that a good understanding @fabien0102 ?

ericadamski: Without assuming the structure of the base url I cannot think of a way of using the props that exist. We need some way of knowing what the expected default is.

We can be opinionated and assume that a base url is exactly: `http(s)://[url-ish]/`

We are equating the route host with the base url prop.

Is that the correct assumption?

fabien0102: Since `RestfulReactProvider` and `Get` derived from `React.Context`, you can simply store the `base` and the `nestedBase` inside the context and consume these both variables without assuming anything.

A think that you should study a bit more the actual structure of the app ;) 
https://reactjs.org/docs/context.html

src/Get.test.tsx
fabien0102: It can works! never try with `nock.done()` but it's actually a very good idea 👍

src/Context.tsx
ericadamski: Sorry @fabien0102 I may have been worried about adding more props to the interface. The new push should have resolved this. It adds the original base to the context

src/Get.tsx
```diff
@@ -214,7 +232,7 @@ function Get<TData = any, TError = any>(props: GetProps<TData, TError>) {
   return (
     <RestfulReactConsumer>
       {contextProps => (
-        <RestfulReactProvider {...contextProps} base={`${contextProps.base}${props.path}`}>
+        <RestfulReactProvider {...contextProps} base={`${contextProps.base}/${props.path}`}>
```
peterszerzo: Should concatenation happen here regardless of whether the `path` prop has a leading slash or not? I wonder how relativeness is reconciled further down the component tree.

ericadamski: Could you expand more @peterszerzo ? If we have no leading slash and concat it would look like `https://example.complop`. Maybe you are saying if I want to do a relative path but started out with a parent like `/plop` then the concatenation would be `https://example.com//plop`.

I may be missing a normalization of the path here.

src/Context.tsx
```diff
@@ -4,6 +4,8 @@ import { ResolveFunction } from "./Get";
 export interface RestfulReactProviderProps<T = any> {
   /** The backend URL where the RESTful resources live. */
   base: string;
+  /** The original backend URL where the RESTful resources live. */
+  originalBase?: string;
```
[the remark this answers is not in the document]
ericadamski: It doesn't get changed over time. It is the original base passed in regardless of nesting.

src/Context.tsx
```diff
@@ -21,10 +23,20 @@ const { Provider, Consumer: RestfulReactConsumer } = React.createContext<Restful
   requestOptions: {},
 });
 
+let originalBase: string = "";
```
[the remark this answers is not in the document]
ericadamski: Storing this so that the `originalBase` prop can not get overridden by using input. It is set internally by the constructor and passed in through contexted

src/Get.tsx
peterszerzo: What seems more intuitive to me is that if the `path` prop starts with a leading slash, the `base` prop of the context provider stays unmodified. This would allow child components to work with their grandparents' `base` prop, essentially dropping whatever the parent has set (as opposed to picking up on it in case the `path` didn't start with a slash).

ericadamski: I am still not seeing the absolute path escape through a method like this. I must fully be overlooking some context here.

if the base url is `https://example.com`

```jsx
<Get path="/absolute"> // fetch 'https://example.com/absolute'
  <Get path="relative" > // fetch 'https://example.com/relative' (because the absolute didn't get appended)
    <Get path="again" > // fetch 'https://example.com/relative/again'
      // how do we know what the base is if the base got modified?
      // What absolute path do I fetch here? At this point base
      // would be https://exapmle.com/relative/again
       <Get path="/more-absolute" />
   </Get>
 </Get>
</Get>
```

ericadamski:
Thank you @fabien0102 @peterszerzo @TejasQ for spending so much time here. I am relatively embarrassed about taking up so much of your time 🤦‍♂️ and still not getting out what you want.

src/Get.tsx
ericadamski: I see! I misunderstood. I thought you were saying concatenating was the solution without the `originalBase`. My mistake.

The semantics around fetching here are complex. My opinion would be that nested routes are always relative regardless of leading slash. If you wanted to make an internal route fetch escape to the base URL then a prop of absolute might make it more clear?

```jsx
<Get path="/absolute"> // fetch 'https://example.com/absolute'
  <Get path="relative" > // fetch 'https://example.com/absolute/relative'
    <Get path="again" > // fetch 'https://example.com/absolute/relative/again'
       // This would hit the originalBase
       <Get absolute path="more-absolute" />
   </Get>
 </Get>
</Get>
```

But I do not have enough context on the usage patterns of this lib I am not sure what the implicit interpretation of the routes with leading slashes is in the "wild" 😆

peterszerzo: Awesome, I think this is a lot better - the presence or absence of a slash is too little visual noise for difference in behavior so significant. I would actually be in favor of recommending slash-less `path` throughout the usage of the library (whether through warnings or just phrasing all examples that way).

peterszerzo:
@ericadamski it's really ok - I am enjoying the bouncing off ideas very much so far :)

TejasQ:
@ericadamski, no problem at all. We enjoy collaboration with great folks – you have no reason whatsoever to be emarrased.

ericadamski:
Should `Mutate` and `Poll` follow the same nesting logic as `Get`?

README.md
```diff
@@ -143,19 +143,25 @@ import Get from "restful-react";
...
           <h1>Random Image</h1>
           {/* Composes path with parent: sends request to /breeds/image/random */}
-          <Get path="/image/random">
+          <Get path="image/random">
+            {image => <img alt="Random Image" src={image && image.message} />}
+          </Get>
+
+          <h1>Random Breed</h1>
+          {/* Escapes the relative path with parent: sends request to /random */}
+          <Get absolute path="random">
```
ericadamski: This is the only documented place for the `absolute` prop, should we add a more detailed description to emphasize its importance?

README.md
```diff
@@ -169,6 +175,8 @@ export default () => (
 
 From the above example, _not only_ does the path accumulate based on the nesting of each `Get`, but each `Get` _can_ override its parent with other props as well: including having _specific_ `requestOptions` if there was a valid use case.
 
+To escape form nested path accumulation you can sepcify the `absolute` prop which will ignore all parent relative paths.
```
ericadamski: Typo here, I will fix it :)

README.md
```diff
@@ -320,7 +328,7 @@ RESTful React also exports a `Poll` component that will poll a backend endpoint
 ```jsx
 import { Poll } from "restful-react"
 
-<Poll path="/deployLogs" resolve={data => data && data.data}>
+<Poll path="deployLogs" resolve={data => data && data.data}>
```
ericadamski: I removed the leading `/`s here but I am not sure the code will support this change. See https://github.com/contiamo/restful-react/pull/33#issuecomment-417657956

src/Context.tsx
```diff
@@ -13,18 +13,23 @@ export interface RestfulReactProviderProps<T = any> {
...
   requestOptions?: (() => Partial<RequestInit>) | Partial<RequestInit>;
+  /**
+   * The initial base path given to the parent.
+   */
+  originalBase?: string;
 }
 
 const { Provider, Consumer: RestfulReactConsumer } = React.createContext<RestfulReactProviderProps>({
   base: "",
+  originalBase: undefined,
```
ericadamski: I made this `undefined` to make the existence check syntactically simpler, no other reason than that.

src/Get.test.tsx
```diff
@@ -172,6 +229,34 @@ describe("Get", () => {
     });
   });
 
+  describe("with absolute", () => {
```
ericadamski: This test is pretty much a duplicate of a test above for composing routes. It is more explicit what it is testing thought. Just pointing it out 😉

src/Get.tsx
```diff
@@ -214,8 +238,17 @@ function Get<TData = any, TError = any>(props: GetProps<TData, TError>) {
...
-        <RestfulReactProvider {...contextProps} base={`${contextProps.base}${props.path}`}>
-          <ContextlessGet {...contextProps} {...props} />
+        <RestfulReactProvider
+          {...contextProps}
+          base={props.absolute ? contextProps.base : `${contextProps.base}/${normalizeUrlPath(props.path)}`}
+        >
+          <ContextlessGet
+            {...contextProps}
+            {...props}
+            originalBase={
```
ericadamski: This statement deals with the manual overriding of the base prop. See [this test](https://github.com/contiamo/restful-react/pull/33/files#diff-ecba994a57db0df28ccfaccfaba2a1e0R280) that covers it.

README.md
TejasQ: In the props interface. :wink:

README.md
ericadamski: I am not sure that Poll should support the nested relative route structure. It could be huge resource hog when nesting multiple. thoughts? @peterszerzo @TejasQ @fabien0102

TejasQ: Polls would never be nested, but should have a similar API to the other components in order to not confuse users.

src/Context.tsx
TejasQ: This is a comment you'd leave in code since you're the author, as opposed to on GitHub where it goes away with the PR. 🙂

src/Get.test.tsx
TejasQ: [Maybe describe the intent in the code as well](https://github.com/contiamo/restful-react/pull/33#discussion_r214679459).

README.md
ericadamski: Just to be clear, you are talking specifically about the leading slashes of the `path`? @TejasQ

TejasQ: Yes.

peterszerzo:
I have the feeling that we might be better off if we designated this as a learning PR and came back to it with fresh eyes a bit later. I realize that we went through a few ideas on how to go about this feature, and unfortunately it is still not clear to me which style of API or implementation would serve us best. I take that as a sign that should gather more insight while using `restful-react` on larger production apps at work, until we have a better handle on which road we go down on. What do we think?

ericadamski:
I would agree @peterszerzo, I like the idea of having feedback from real use cases to determine what the issues actually are. That being said I do not have enough context to really give a solid opinion 😄

peterszerzo:
Closing this for now then - thank you all for your hard work and helpful discussion. We'll reference this PR once the new one opens up sometime in the coming months.

PR #41, opened 2018-09-10
Fix return type

fabien0102:
# Why

I missed this little return type for the mutate return type during my last PR 😄 

This PR fix this little mistake


TejasQ:
Beautiful.

PR #43, opened 2018-09-12
Add a clean way to consolidate data in `Poll`

fabien0102:
# Why

Our server send us chunk instead of a complete response on long polling (for obvious performance reasons ^^)

The idea is to have `prevData` available in the `Poll:resolve` props, so we can do:

```jsx
<Poll path="" resolve={(data, prevData) => (prevData || "") + data}>
  {/* my awesome application */}
</Poll>
```

I also add a lot of unit tests to `Poll` and fix some little behaviour on my way:
 -  avoid a useless call on start
 - cancel the request on component did mount

### DIFF
--- a/src/Poll.tsx
+++ b/src/Poll.tsx
@@ -95,7 +95,7 @@ export interface PollProps<TData, TError> {
   /**
    * Should the data be transformed in any way?
    */
-  resolve?: GetProps<TData, TError>["resolve"];
+  resolve?: (data: any, prevData: TData) => TData;
   /**
    * We can request foreign URLs with this prop.
    */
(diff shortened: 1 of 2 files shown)


src/Poll.tsx
```diff
@@ -199,54 +216,70 @@ class ContextlessPoll<TData, TError> extends React.Component<
...
+  public stop = () => {
     this.keepPolling = false;
     this.setState(() => ({ polling: false, finished: true })); // let everyone know we're done here.
   };
 
   public componentDidMount() {
     const { path, lazy } = this.props;
 
-    if (!path) {
+    if (path === undefined) {
```
TejasQ: But `path` is required in the interface 🤔

fabien0102: Really? Why can we not poll on the `base`? It's works for `Get` so…

fabien0102: I keep the `path` mandatory to avoid mistake, so we need to `path=""` if we want this case (a bit stupid but explicit)

fabien0102: Ah, I may have your comment, it's for javascript users ^^ (this case is not possible in typescript of course)

TejasQ: Makes sense.

PR #46, opened 2018-09-12
Improve the error management for Poll

fabien0102:
# Why

Because as a noob I've commited a `.skip` on this part 😆 and also because we have a lot of foggy spots on this part.

The idea is to be as close as possible of `Get` regarding the comportement and also continue polling after an error (so the `until` props is really useful)

### DIFF
--- a/src/Poll.test.tsx
+++ b/src/Poll.test.tsx
@@ -196,7 +196,7 @@ describe("Poll", () => {
     });
   });
 
-  describe.skip("with error", () => {
+  describe("with error", () => {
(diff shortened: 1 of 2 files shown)


TejasQ:
You've got to be kidding me. I should have caught this in the review!

TejasQ:
Using yarn as the package manager for this project is a real problem. Our `.prettierrc`s are not synced because of https://github.com/yarnpkg/yarn/issues/5476, since the postinstall from https://github.com/contiamo/operational-scripts doesn't execute. :(

PR #54, opened 2018-09-21
resolve urls instead concat url strings

chilijung:
# Why

URLs in methods are resolving unexpected, due to the urls are simply concatenating. This PR use `url.resolve` to enable users to use absolute and relative paths expectedly.

### DIFF
--- a/src/Get.tsx
+++ b/src/Get.tsx
@@ -1,6 +1,7 @@
 import { DebounceSettings } from "lodash";
 import debounce from "lodash/debounce";
 import * as React from "react";
+import url from "url";
 import RestfulReactProvider, { InjectedProps, RestfulReactConsumer, RestfulReactProviderProps } from "./Context";
 import { processResponse } from "./util/processResponse";
 
@@ -89,7 +90,7 @@ export interface GetProps<TData, TError> {
    * to fetch from an entirely different URL.
    *
    */
-  base?: string;
+  base: string;
   /**
    * How long do we wait between subsequent requests?
    * Uses [lodash's debounce](https://lodash.com/docs/4.17.10#debounce) under the hood.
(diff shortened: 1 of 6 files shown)


src/Get.tsx
```diff
@@ -89,7 +90,7 @@ export interface GetProps<TData, TError> {
    * to fetch from an entirely different URL.
    *
    */
-  base?: string;
+  base: string;
```
fabien0102: This props needs to be optional if we want to use `<Get />` with specify any `base`, since you have a defaultProps to always have a value, you can simply use the typescript bang operator:


```typescript
this.props.base! // adding ! tell to typescript that this this.props.base can't be undefined
```

A very nice article about this -> https://levelup.gitconnected.com/ultimate-react-component-patterns-with-typescript-2-8-82990c516935

(and normally the last version of typescript can deal with this defaultProps case, but I don't try yet)

src/Mutate.tsx
```diff
@@ -33,7 +34,7 @@ export interface MutateCommonProps {
    * The path at which to request data,
    * typically composed by parents or the RestfulProvider.
    */
-  path?: string;
+  path: string;
```
fabien0102: Same as my previous comment, we need to keep all these props optionals

chilijung:
Thanks for the tips! Updated!

fabien0102:
You have forgot the `Mutate` component :wink:

fabien0102:
All green! :tada: Last thing for me, can you add some little tests to show what cases this `url.resolve`  solve? :angel:

fabien0102:
@TejasQ We have this error on Travis:

```sh
Error: Cannot use authenticated API requests.
```

It remember me a past issue, but I don't remember how to fix this ^^ Can you have a look?

@chilijung Don't worry about this Travis, all it's green 💯

TejasQ:
Fixed! If this is rebased on the master, travis should be happy and we can merge 🚀

PR #55, opened 2018-09-21
Remove engine in package.json

chilijung:
# Why

I don't think the `engine` setting is right in `package.json` 

It cause built fail in now

```
> warning @canner/restful-react@5.1.1: The engine "vscode" appears to be invalid.
> info fsevents@1.2.4: The platform "linux" is incompatible with this module.
> info "fsevents@1.2.4" is an optional dependency and failed compatibility check. Excluding it from installation.
> warning restful-react@5.2.0: The engine "vscode" appears to be invalid.
> [4/5] Linking dependencies..
```

FYI: https://docs.npmjs.com/files/package.json#engines

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,9 +7,6 @@
   "files": [
     "lib"
   ],
-  "engines": {
-    "vscode": "^1.22.0"
-  },
   "keywords": [
     "rest",
     "restful",


fabien0102:
Logically it's just a warning, but I'm agree that this vscode as engine is weird 😄

chilijung:
In `now` it will break the build process.

TejasQ:
Thank you, @chilijung!

Issue #29, opened 2018-08-22
Improve clarity of backend route composition


peterszerzo:
**Is your feature request related to a problem? Please describe.**
If I start out a page with the following markup:
```
<Get path="/people">
  {data => (
    <Get path="/products" />
  )}
</Get>
```
I find it unintuitive that the resulting backend path in the child `Get` component is `${base}/people/products`, which, in an HTML analogy, would make more sense if I omitted the slash like in `<Get path="products">`. It is an overhead to have to do `../products` in order to replicate the 'absolute path' behavior, or to pull in the backend URL again to set as `base` prop. I wonder if we can make this more predictable.

**Describe the solution you'd like**
The following code:
```
<Get path="/people">
  {data => (
    <Get path="/products" />
  )}
</Get>
```
should make requests to `${base}/people` and `${base}/products`, whereas 
```
<Get path="/people">
  {data => (
    <Get path="products" />
  )}
</Get>
```
should make requests to `${base}/people` and `${base}/people/products`.

In this proposal, URL composition can be disabled by simply adding a slash. As a personal preference, I can even see myself forcing absolute paths at all times in order to remove coupling my components to the entire component hierarchy (and thereby making rearranging components risky). We could even enforce this as a kind of strict mode, or issue runtime warnings if the user opts into them.

**Alternatives**
I can foresee some confusion happening when relative and absolute positions are nested, but I'm sure we can figure out a nice set of rules on the implementation level. Maybe we can gather some data from projects using `restful-react` - how much nesting/composition is desired and how nesting complexity is managed - to see what API makes the most sense.

TejasQ:
This is so awesome. Let's do it.

ericadamski:
@TejasQ @peterszerzo I would love to help out with this issue!

TejasQ:
Do it.

ericadamski:
There are a couple directions or question about the implementation here that @peterszerzo suggested, I just want to lay out my understanding of questions going forward and next steps.

The main goal is to add relative paths based on the nested `<Get />` calls to match the perceived interaction @peterszerzo mentioned.

> Describe the solution you'd like
> The following code:
>
> ```
>  <Get path="/people">
>   {data => (
>      <Get path="/products" />
>    )}
>  </Get>
> ```
>
> should make requests to ${base}/people and ${base}/products, whereas
>
> ```
> <Get path="/people">
>  {data => (
>    <Get path="products" />
>  )}
> </Get>
>
[...]

PR #64, opened 2018-10-05
compose urls for relative paths cases

elangobharathi:
# Why

To compose relative urls explicitly using no forward slash in path value. Refer #29 for more background.

# Outcome

If the path starts with forward slash, it is considered as absolute url. If not, it is considered as relative url.
For example,
```
base = "/people" and path = "/absolute" resolves to "/absolute"
```
whereas,
```
base = "/people" and path = "relative" resolves to "/people/relative"
```

# Note

In the nested cases, the immediate parent's url becomes base for composition.
For example, 
```
 <RestfulProvider base="https://my-awesome-api.fake/MY_SUBROUTE">
  <Get path="/absolute-1">
    {() => (
      <Get path="relative-1">
        {() => (
          <Get path="/absolute-2">
            {() => <Get path="relative-2">{() => <Get path="relative-3">{children}</Get>}</Get>}
          </Get>
        )}
      </Get>
    )}
  </Get>
</RestfulProvider>
```
emits 
`"https://my-awesome-api.fake/MY_SUBROUTE/absolute-1"`
`"https://my-awesome-api.fake/MY_SUBROUTE/absolute-1/relative-1"`
`"https://my-awesome-api.fake/MY_SUBROUTE/absolute-2"`
`"https://my-awesome-api.fake/MY_SUBROUTE/absolute-2/relative-2"`
`"https://my-awesome-api.fake/MY_SUBROUTE/absolute-2/relative-2/relative-3"`

### DIFF
--- a/src/Get.test.tsx
+++ b/src/Get.test.tsx
@@ -48,23 +48,6 @@ describe("Get", () => {
       await wait(() => expect(children.mock.calls.length).toBe(2));
     });
 
-    it("should compose the url with the base", async () => {
-      nock("https://my-awesome-api.fake")
-        .get("/plop")
-        .reply(200);
-
-      const children = jest.fn();
-      children.mockReturnValue(<div />);
-
-      render(
-        <RestfulProvider base="https://my-awesome-api.fake">
-          <Get path="/plop">{children}</Get>
-        </RestfulProvider>,
-      );
-
-      await wait(() => expect(children.mock.calls.length).toBe(2));
-    });
-
     it("should set loading to `true` on mount", async () => {
       nock("https://my-awesome-api.fake")
         .get("/")
@@ -163,9 +146,9 @@ describe("Get", () => {
       expect(children.mock.calls[1][0]).toEqual(null);
       expect(children.mock.calls[1][1].error).toEqual({
         data:
-          "invalid json response body at https://my-awesome-api.fake/ reason: Unexpected token < in JSON at position 0",
+          "invalid json response body at https://my-awesome-api.fake reason: Unexpected token < in JSON at position 0",
         message:
-          "Failed to fetch: 200 OK - invalid json response body at https://my-awesome-api.fake/ reason: Unexpected token < in JSON at position 0",
+          "Failed to fetch: 200 OK - invalid json response body at https://my-awesome-api.fake reason: Unexpected token < in JSON at position 0",
       });
     });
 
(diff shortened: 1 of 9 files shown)


src/Get.tsx
```diff
@@ -200,10 +200,25 @@ class ContextlessGet<TData, TError> extends React.Component<
...
+     * If the path starts with slash, it is considered as absolute url.
+     * If not, it is considered as relative url.
+     * For example,
+     * base = "http://example.com/people" and path = "/absolute"
+     * reolves to "http://example.com/absolute"
+     * whereas,
+     * base = "http://example.com/people" and path = "relative"
+     * reolves to "http://example.com/people/relative"
+     */
+    const composedUrl = requestPath
```
TejasQ: rad

This set of nested ternaries is so rad, but maybe we can refactor this a bit in order to preserve readability so that when we look at it in a few months we'll understand it faster.

src/Mutate.tsx
```diff
@@ -122,8 +122,12 @@ class ContextlessMutate<TData, TError> extends React.Component<
 
     const requestPath =
       verb === "DELETE" && typeof body === "string"
-        ? url.resolve(base!, url.resolve(path, body))
-        : url.resolve(base!, path);
+        ? path.startsWith("/")
```
TejasQ: See the other comment about nested ternaries. This also appears to be duplicated logic so maybe put it in a function that is unit tested?

src/Mutate.tsx
```diff
@@ -120,10 +121,12 @@ class ContextlessMutate<TData, TError> extends React.Component<
     const { base, path, verb, requestOptions: providerRequestOptions } = this.props;
     this.setState(() => ({ error: null, loading: true }));
 
-    const requestPath =
-      verb === "DELETE" && typeof body === "string"
-        ? url.resolve(base!, url.resolve(path, body))
-        : url.resolve(base!, path);
+    let requestPath: string;
```
fabien0102: I think that you can keep the ternary here, for 2 choices is fine 😃

src/Mutate.tsx
```diff
@@ -181,7 +184,14 @@ function Mutate<TError = any, TData = any>(props: MutateProps<TData, TError>) {
   return (
     <RestfulReactConsumer>
       {contextProps => (
-        <RestfulReactProvider {...contextProps} base={`${contextProps.base}${props.path}`}>
+        <RestfulReactProvider
+          {...contextProps}
+          base={
+            props.path.startsWith("/")
```
fabien0102: I don't really check how `url.resolve` works, but this should do the stuff for the both cases.

src/util/composeUrl.test.ts
```diff
@@ -0,0 +1,35 @@
+import { cleanup } from "react-testing-library";
+import url from "url";
+import { composeUrl } from "./composeUrl";
+
+afterEach(() => {
+  cleanup();
```
fabien0102: Don't need to cleanup here, we don't have any react implementation

src/util/composeUrl.test.ts
```diff
@@ -0,0 +1,35 @@
...
+});
+
+describe("compose url", () => {
+  it("should provide absolute url", () => {
+    const base = "https://my-awesome-api.fake/people";
```
fabien0102: Can you add a test with a trailing slash?

src/Poll.tsx
```diff
@@ -209,7 +210,7 @@ class ContextlessPoll<TData, TError> extends React.Component<
     const { lastPollIndex } = this.state;
     const requestOptions = this.getRequestOptions();
 
-    const request = new Request(url.resolve(base!, path), {
+    const request = new Request(path.startsWith("/") ? url.resolve(base!, path) : `${base!}/${path}`, {
```
fabien0102: This can be replace by `composeUrl`

src/Get.tsx
```diff
@@ -260,7 +277,14 @@ function Get<TData = any, TError = any>(props: GetProps<TData, TError>) {
   return (
     <RestfulReactConsumer>
       {contextProps => (
-        <RestfulReactProvider {...contextProps} base={url.resolve(contextProps.base, props.path)}>
+        <RestfulReactProvider
+          {...contextProps}
+          base={
+            props.path.startsWith("/")
```
fabien0102: This can be replace by `composeUrl`

src/Get.test.tsx
```diff
@@ -65,6 +65,89 @@ describe("Get", () => {
...
+                  <Get path="/absolute-2">
+                    {() => <Get path="relative-2">{() => <Get path="relative-3">{children}</Get>}</Get>}
+                  </Get>
+                )}
+              </Get>
+            )}
+          </Get>
+        </RestfulProvider>,
+      );
+      await wait(() => expect(children.mock.calls.length).toBe(6));
```
fabien0102: Can you also add some other expects with some data from nock to be sure that we have that we are expected here? I want to know if the `.get("/absolute-2/relative-2/relative-3")` is well called

fabien0102:
After checking how `url.resolve` works a bit closer, you are just lucky in your test cases.

```js
url.resolve('http://test.com/one/two/three', '/four'); // http://test.com/four
url.resolve('http://test.com/one/two/three', 'four'): // http://test.com/one/two/three/four
```

I was wondering how it was working, and now I finally understand 😄 

To provide a real implementation of this feature, you need to add some tests with some `base` more complicated like this:

```jsx
render(
    <RestfulProvider base="https://my-awesome-api.fake/absolute-1">
      <Get path="">
        {() => (
          <Get path="relative-1">
            {() => (
              <Get path="/absolute-2">
                {() => <Get path="relative-2">{() => <Get path="relative-3">{children}</Get>}</Get>}
              </Get>
            )}
          </Get>
        )}
      </Get>
    </RestfulProvider>
)
```

and also some trailing slashes. I think that `url.resolve` can be a very helping piece in this feature, you "just" need to store the `rootBase` somewhere.

TejasQ:
Great catch, @fabien0102!

@elangobharathi, apologies, we may have not been as clear in describing the issue: the idea is that the `base` prop is the _absolute root_ of any nested `Get` components whose `path`s are prefixed with a `/`, even if the `base` prop itself contains subpaths.

The current implementation always rolls up to the domain-level instead of a subpath on the `base` prop.

micha-f:
I found this a while ago - https://github.com/Canner/restful-react/commit/978a686a14c39d9b6c7b65ee4729a520e8b2223b 

These changes were only made in their local branch.

elangobharathi:
@TejasQ I am sorry I couldn't fully understand. Could you please explain me this with an example?

> @elangobharathi, apologies, we may have not been as clear in describing the issue: the idea is that the `base` prop is the _absolute root_ of any nested `Get` components whose `path`s are prefixed with a `/`, even if the `base` prop itself contains subpaths.
> 
> The current implementation always rolls up to the domain-level instead of a subpath on the `base` prop.

elangobharathi:
@fabien0102 Thank you for the comments :) It would be great if you can provide the expected result for your example test case. This would help me understand the expectations a bit more.

> ```js-jsx
> render(
>     <RestfulProvider base="https://my-awesome-api.fake/absolute-1">
>       <Get path="">
>         {() => (
>           <Get path="relative-1">
>             {() => (
>               <Get path="/absolute-2">
>                 {() => <Get path="relative-2">{() => <Get path="relative-3">{children}</Get>}</Get>}
>               </Get>
>             )}
>           </Get>
>         )}
>       </Get>
>     </RestfulProvider>
> )
> ```

TejasQ:
Absolutely!

Here, let's use your PR description with some adjustments in order to make it a little more understandable:

> If the path starts with forward slash, it is considered as absolute url. If not, it is considered as relative url.

^ this is correct.

>For example,
> ```
> base = "http://example.com/people" and path = "/absolute" 
> resolves to "http://example.com/absolute"
> ```

^ this is incorrect.

Instead,
```
base = "http://example.com/people" and path = "/absolute" 
resolves to "http://example.com/people/absolute"

because base is /people!
```

> whereas,
> ```
> base = "http://example.com/people" and path = "relative" 
> resolves to "http://example.com/people/relative"
> ```

^ Correct.

# Note

In the nested cases, the immediate parent's url becomes base for composition.
For example, 
```
 <RestfulProvider base="https://my-awesome-api.fake/MY_SUBROUTE">
  <Get path="/absolute-1">
    {() => (
      <Get path="relative-1">
        {() => (
          <Get path="/absolute-2">
            {() => <Get path="relative-2">{() => <Get path="/not-relative-3">{children}</Get>}</Get>}
          </Get>
        )}
      </Get>
    )}
  </Get>
</RestfulProvider>
```
emits 
`"https://my-awesome-api.fake/MY_SUBROUTE/absolute-1"`
`"https://my-awesome-api.fake/MY_SUBROUTE/absolute-1/relative-1"`
`"https://my-awesome-api.fake/MY_SUBROUTE/absolute-2"`
`"https://my-awesome-api.fake/MY_SUBROUTE/absolute-2/relative-2"`
`"https://my-awesome-api.fake/MY_SUBROUTE/not-relative-3"`

Does that clear things up? Let me know if I can clarify further. 😄

fabien0102:
@TejasQ was quicker than me 🥇 @elangobharathi is it clear for you or do you want some test cases?

elangobharathi:
@TejasQ That cleared up the base with subpath case `https://my-awesome-api.fake/MY_SUBROUTE/`. Thank you.

@fabien0102 @TejasQ I would definitely need the expected result for some complicated nesting cases.

Can you provide the expected result for the following cases?

```
render(
    <RestfulProvider base="https://my-awesome-api.fake/absolute-1">
      <Get path="">
        {() => (
          <Get path="relative-1">
            {() => (
              <Get path="/absolute-2">
                {() => <Get path="relative-2">{() => <Get path="relative-3">{children}</Get>}</Get>}
              </Get>
            )}
          </Get>
        )}
      </Get>
    </RestfulProvider>
)
```

```
render(
    <RestfulProvider base="https://my-awesome-api.fake/absolute-1">
      <Get path="/">
        {() => (
          <Get path="relative-1">
            {() => (
              <Get path="">
                {() => <Get path="relative-2">{() => <Get path="relative-3">{children}</Get>}</Get>}
              </Get>
            )}
          </Get>
        )}
      </Get>
    </RestfulProvider>
)
```

fabien0102:
If I don't do any mistake, this should result like this:
```tsx
    it("should compose more nested absolute and relative urls", async () => {
      nock("https://my-awesome-api.fake")
        .get("/absolute-1")
        .reply(200);

      nock("https://my-awesome-api.fake")
        .get("/absolute-1/relative-1")
        .reply(200);

      nock("https://my-awesome-api.fake")
        .get("absolute-1/absolute-2")
        .reply(200);

      nock("https://my-awesome-api.fake")
        .get("absolute-1/absolute-2/relative-2")
        .reply(200);

      nock("https://my-awesome-api.fake")
        .get("absolute-1/absolute-2/relative-2/relative-3")
        .reply(200, { path: "absolute-1/absolute-2/relative-2/relative-3" });

      const children = jest.fn();
      children.mockReturnValue(<div />);

      render(
        <RestfulProvider base="https://my-awesome-api.fake/absolute-1">
          <Get path="">
            {() => (
              <Get path="relative-1">
                {() => (
                  <Get path="/absolute-2">
                    {() => <Get path="relative-2">{() => <Get path="relative-3">{children}</Get>}</Get>}
                  </Get>
                )}
              </Get>
            )}
          </Get>
        </RestfulProvider>,
      );
      await wait(() => expect(children.mock.calls.length).toBe(6));
      expect(children.mock.calls[5][0]).toEqual({ path: "absolute-1/absolute-2/relative-2/relative-3" });
    });
```

And same for `Get[0].path="/"`  or `RestfulProvider.base = "https://my-awesome-api.fake/absolute-1/"`

elangobharathi:
@fabien0102 Thank you for the details :blush:.

@TejasQ I am waiting for your opinion.
>
>Note: I'm not totally sure abouth the /absolute2 composition, let me ask to @TejasQ to have a >second opinion 😉
>

TejasQ:
He asked me! He is 💯 % correct. :)

src/Mutate.tsx
elangobharathi: Changed it back to ternary :smiley:

elangobharathi:
@fabien0102 @TejasQ Thank you for your comments. I have addressed them now. Please review and let me know your feedback :blush:

src/Poll.tsx
elangobharathi: Yes, replaced it now. Thanks

src/Get.tsx
elangobharathi: Yes, replaced it now. Thanks

src/Get.test.tsx
```diff
@@ -40,31 +40,14 @@ describe("Get", () => {
...
-        <RestfulProvider base="https://my-awesome-api.fake/">
+        <RestfulProvider baseRoot="https://my-awesome-api.fake/">
           <Get path="/">{children}</Get>
         </RestfulProvider>,
       );
 
       await wait(() => expect(children.mock.calls.length).toBe(2));
     });
 
-    it("should compose the url with the base", async () => {
```
TejasQ: Please do not remove test cases, or if you do, please say _why_.

elangobharathi:
Hi @TejasQ Yes, you are right. We should definitely rename them. How about naming them as `base` and `parentPath`?

For example,
```
<RestfulProvider baseRoot="https://my-awesome-api.fake/MY_SUBROUTE">
  <Get path="/absolute-1">
    {() => (
      <Get path="relative-1">
      </Get>
    )}
  </Get>
</RestfulProvider>
```

How about 
`https://my-awesome-api.fake/MY_SUBROUTE` remains base
`/absolute-1` as parenthPath
`relative-1` remains path

Please let me know if you can think of any better names as well :smile:

TejasQ:
I love that naming! Great idea!

elangobharathi:
@fabien0102 @TejasQ  Thank you for the approvals :blush:

TejasQ:
So @elangobharathi if you can rebase it, we're good to go! WOOOO! THANK YOU FOR YOUR CONTRIBUTION! 🚀 

woo

elangobharathi:
@TejasQ Thank you! I have just rebased it. Please check and merge. :blush:

TejasQ:
Thanks, Elango! It was a true pleasure working with you!

elangobharathi:
Thanks @TejasQ ! It was a pleasure working with you too!

PR #72, opened 2018-10-16
**Feature:** Add support for validators that are promise-based or throw errors

### LINKED ISSUE #57: Allow for validation failures in the `resolve` prop
**Is your feature request related to a problem? Please describe.**
Currently, the `resolve` prop of the `Get` component is a simple function that does some transformation on the data coming from the backend. The developer using `restful-react` is responsible that it is correct (no runtime errors), and add additional logic on their own to handle unexpected responses from the server. In order to offer some legwork to help folks make absolutely sure that the data passing this function either fails gracefully or passes through with sound data, I'd like to suggest two features:
* catch runtime errors in the `fetch` method (adding a `.catch` in https://github.com/contiamo/restful-react/blob/master/src/Get.tsx#L152), sending runtime errors to either the global error management of the library, or the local one if `localErrorsOnly` is enabled. This would help with third-party parsing/validating solutions like https://github.com/jquense/yup#mixedcastvalue-any-any, which are super powerful, but always throw runtime errors when they fail.
* allow for the `resolve` prop to accept a promise which either resolves with the correctly shaped data or rejects with an error. Again, this might help with third-party validation solutions which offer some validation features only asynchronously (recursive data structures and the like).

**Describe alternatives you've considered**
I cannot think of an alternative that doesn't bypass `resolve` entirely, leaving validation to be performed inside the `render` method and killing performance.

**Additional context**
At a high-level, the purpose of these suggestions is to allow a style of programming that does super-strict validation before data reaches the internals of a react component tree. This provides both safety and debugging convenience: there is always a failure before data even reaches presentational views, narrowing down issues to the validator logic, which can be unit-tested with an ever-broadening list of fixtures from the backend. No more looking around `React` code to fix backend-frontend miscommunication.

peterszerzo:
# Why

This PR fixes #57, adding support for validators that either throw runtime errors or work through a promise, extending the support for libraries like `yup`.

### Errors

The errors coming from this validation are standard and can produce anything, so they don't fit into the `{ message: string, data: TError | string }` interface - so as a first pass, I am passing a `RESOLVE_ERROR` string as message so it is easily recognizable, and stringifying the actual error inside `data` - which the end user would have to parse back into JSON later. Wonder if there is a better way to deal with this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -93,6 +93,9 @@ typings/
 public
 lib
 
+# Cache for parcel serverl in example
+.cache
+
 # Files managed by operational-scripts
 tslint.json
 tsconfig.json
--- a/example/README.md
+++ b/example/README.md
@@ -1,0 +1,1 @@
+This is a simple live demo of `restul-react` components primarily used to test features. Run it with `npm run example`.
--- a/src/Get.test.tsx
+++ b/src/Get.test.tsx
@@ -253,6 +253,68 @@ describe("Get", () => {
       await wait(() => expect(children.mock.calls.length).toBe(2));
       expect(children.mock.calls[1][0]).toEqual({ hello: "world", foo: "bar" });
     });
+
+    it("should transform data with a promise", async () => {
+      nock("https://my-awesome-api.fake")
+        .get("/")
+        .reply(200, { hello: "world" });
+
+      const children = jest.fn();
+      children.mockReturnValue(<div />);
+
+      render(
+        <RestfulProvider base="https://my-awesome-api.fake">
+          <Get path="" resolve={data => Promise.resolve({ ...data, foo: "bar" })}>
+            {children}
+          </Get>
+        </RestfulProvider>,
+      );
+
+      await wait(() => expect(children.mock.calls.length).toBe(2));
+      expect(children.mock.calls[1][0]).toEqual({ hello: "world", foo: "bar" });
+    });
+
+    it("should pass an error when the resolver throws a runtime error", async () => {
+      nock("https://my-awesome-api.fake")
+        .get("/")
+        .reply(200, { hello: "world" });
+
+      const children = jest.fn();
+      children.mockReturnValue(<div />);
+
+      render(
+        <RestfulProvider base="https://my-awesome-api.fake">
+          <Get path="" resolve={data => data.apples.oranges}>
+            {children}
+          </Get>
+        </RestfulProvider>,
+      );
+
+      await wait(() => expect(children.mock.calls.length).toBe(2));
+      expect(children.mock.calls[1][1].error.message).toEqual("RESOLVE_ERROR");
+      expect(children.mock.calls[1][0]).toEqual(null);
+    });
+
+    it("should pass an error when the resolver is a promise that rejects", async () => {
+      nock("https://my-awesome-api.fake")
+        .get("/")
+        .reply(200, { hello: "world" });
+
+      const children = jest.fn();
+      children.mockReturnValue(<div />);
+
+      render(
+        <RestfulProvider base="https://my-awesome-api.fake">
+          <Get path="" resolve={data => Promise.reject("nogood")}>
+            {children}
+          </Get>
+        </RestfulProvider>,
+      );
+
+      await wait(() => expect(children.mock.calls.length).toBe(2));
+      expect(children.mock.calls[1][1].error).toEqual({ message: "RESOLVE_ERROR", data: JSON.stringify("nogood") });
+      expect(children.mock.calls[1][0]).toEqual(null);
+    });
   });
 
   describe("with lazy", () => {
@@ -494,10 +556,8 @@ describe("Get", () => {
           </RestfulProvider>,
         ),
       );
-
       await wait(() => expect(apiCalls).toEqual(1));
     });
-
     it("should call the API only 10 times without debounce", async () => {
       let apiCalls = 0;
       nock("https://my-awesome-api.fake")
@@ -505,12 +565,9 @@ describe("Get", () => {
         .get("/?test=XXX")
         .reply(200, () => ++apiCalls)
         .persist();
-
       const children = jest.fn();
       children.mockReturnValue(<div />);
-
       const resolve = a => a;
-
       /**
        * A new instance of RestfulProvider is created on every rerender.
        * This will create a new resolve function every time forcing Get to
@@ -524,7 +581,6 @@ describe("Get", () => {
           <Get path="?test=1">{children}</Get>
         </RestfulProvider>,
       );
-
       times(10, i =>
         rerender(
           <RestfulProvider base="https://my-awesome-api.fake" resolve={resolve}>
@@ -532,7 +588,6 @@ describe("Get", () => {
           </RestfulProvider>,
         ),
       );
-
       expect(apiCalls).toEqual(10);
     });
   });
@@ -543,23 +598,19 @@ describe("Get", () => {
         .get("/")
         .reply(200, () => ++apiCalls)
         .persist();
-
       const children = jest.fn();
       children.mockReturnValue(<div />);
-
       const resolve = a => a;
       const { rerender } = render(
         <RestfulProvider base="https://my-awesome-api.fake" resolve={resolve}>
           <Get path="">{children}</Get>
         </RestfulProvider>,
       );
-
       rerender(
         <RestfulProvider base="https://my-awesome-api.fake" resolve={resolve}>
           <Get path="">{children}</Get>
         </RestfulProvider>,
       );
-
       expect(apiCalls).toEqual(1);
     });
     it("should refetch when base changes", () => {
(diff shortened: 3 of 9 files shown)


.gitignore
```diff
@@ -93,6 +93,7 @@ typings/
 public
 lib
 
+
```
TejasQ: Please remove this change from this PR.

src/Get.tsx
```diff
@@ -225,7 +225,27 @@ class ContextlessGet<TData, TError> extends React.Component<
...
     }
 
-    this.setState({ loading: false, data: resolve!(data) });
+    let resolvedData: TData | null = null;
+    let resolveError: GetDataError<TError> | null = null;
+
+    try {
+      if (resolve) {
+        const resolvedDataOrPromise: TData | Promise<TData> = resolve(data);
+        resolvedData = (resolvedDataOrPromise as { then?: any }).then
```
TejasQ: What is the reasoning behind this type assertion? Tbh it looks a little messy.

peterszerzo: I wasn't able to get it to type-check any other way - looks like detecting whether something is a promise or a function is not that straightforward to follow by the compiler. Do you have any idea how this could improve?

TejasQ: I think we can pair up on it. I'd love to understand the complexity more.

example/README.md
```diff
@@ -0,0 +1,3 @@
+This is a simple live demo of `restul-react` components primarily used to test features.
+
+Run it by getting a global Parcel by running `npm i -g parcel`, then running `parcel example/index.html` from root.
```
TejasQ: I think we can add parcel as a devDependency and it should be fine.

peterszerzo: Oh right, makes sense :)

src/Get.tsx
peterszerzo: Awesome - I managed to confirm via the parcel example that the new features are working, now it's just a matter of cleaning up the code.

example/README.md
```diff
@@ -0,0 +1 @@
+This is a simple live demo of `restul-react` components primarily used to test features. Run it with `npm run example`.
```
fabien0102: We already have codesandbox example and unit tests. If you can avoid to introduce something else to maintain… And I'm sorry but this is not replacing unit tests. I can help you to add the correct unit tests for this feature if needed but please remove this.

src/Get.test.tsx
```diff
@@ -538,4 +543,25 @@ describe("Get", () => {
...
+
+      render(
+        <RestfulProvider base="https://my-awesome-api.fake">
+          <Get path="" resolve={res => res.a.b.c.d.e.f}>
+            {children}
+          </Get>
+        </RestfulProvider>,
+      );
+
+      await wait(() => expect(children.mock.calls.length).toBe(1));
```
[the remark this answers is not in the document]
TejasQ: Good catch. Might make more sense to make the test a little more expressive.

example/README.md
TejasQ: This is just a `dev-server` kind of like in Operational. It doesn't really add much bloat to the library. It's just an on-device, in-browser player like GraphiQL. Ideally, we don't maintain this but make sure our features work _in a browser context_.

I think the tests:
- should be maintained
- add a _lot_ of value

_but_, there is no guarantee that they work 100% in a browser environment with calls to _real_ APIs. This will serve as a check we do _during development_ to make sure things work before pushing.

I don't really see any major costs to having this around. Do you?

TejasQ: @peterszerzo he does have a point though – we _should_ ideally add unit tests for the `resolve` behavior as well.

peterszerzo: I see this more as an integration test as opposed to a unit test - it makes sure that a real request can be made, goes through the state management, and renders something in the end. Which is something I am not confident a unit test suite would catch, especially after a feature this complicated.

peterszerzo: I don't dispute the need for improving the tests for the new `resolve` - I just don't feel comfortable shipping this feature without validating it with an end-to-end example. This could even grow to be like the visual test cases suite in https://github.com/contiamo/operational-visualizations

fabien0102: I got your point, for now I just see this dev server as an escape to implement real tests… After why not, but if we go with an example folder, I will prefer something that we can deploy on codesandbox with more examples as a real addition to our documentation.

I'm just worry about the "it's works on my little example, let's skip the tests for now" 😉

fabien0102: And at this library stage, I'm not agree with you, **we must have** unit tests for every features. Without tests, we can't be sure that the functionally works as expected, it's hard to maintained and everybody can broke the feature by mistake (and by everybody, I include myself of course).

fabien0102: I'm maybe a bit strict on this stack, but all our product depends on this library :wink:

TejasQ: > I'm just worry about the "it's works on my little example, let's skip the tests for now" 😉

As long as we have you on the team, I am confident that this will never happen. 😉  Let's also look at making the examples richer and exportable to CodeSandbox (thanks, by the way, @CompuIves) at a later stage.

peterszerzo: @fabien0102 we don't have to merge until we have a satisfactory test suite, and that would have been the case whether we had the example server or not. Both methods of testing are good for reliability and for catching issues that can bubble up in dependent projects - there is no reason to jump against one of them.

fabien0102: @peterszerzo I agree that both methods of testing are good in the context of a PR, but our goal is to have a maintainable product in the long term. For this purpose, unit tests ensure a non regression in the future, it's automated in the CI, etc etc... So no, both methods of testing are not equal in this case.

peterszerzo:
The unit tests have been added, testing both validation succeeds and failures in the runtime error and promise cases.

fabien0102:
I have some questions/remarks about this PR:

1. Why are we introduce async into `resolve` function? I will prefer to be stick to a sync function here to avoid missunderstanding of our library (for example using another `fetch` in this `resolve` function).
For validating we can use https://www.npmjs.com/package/yup#mixedvalidatesyncvalue-any-options-object-any so this keep this stack very simple and performant.

2. Can we save our `yarn.lock` and git rebase everything to avoid 2/3 huges commits that make our repository bigger for nothing?

3. I'm not totally against the dev server, but it's not in the scope of this issue/pr, can we open an issue to discuss a bit more about this? :angel:

TejasQ:
> 3. I'm not totally against the dev server, but it's not in the scope of this issue/pr, can we open an issue to discuss a bit more about this? 👼

I'm not sure this needs to be a discussion/rfc. It adds value to this and future PRs by adding an extra layer of validation in a browser environment that are a nice complement to the already great suite of unit tests we have.

Are there concrete concerns about the dev server that you have? Maybe we can have a shorter discussion about them here. 🙂 

I get that it's not the scope of this PR, but it _does_ help us test this PR and so for that reason, I think it's quite alright to leave it in here.

peterszerzo:
@fabien0102 some thoughts on your comments:
1. async makes things cleaner in some cases and necessary in others when lazy/recursive and transformation is involved. I am happy to share my use-cases to see if we can get away without this.
2. sure.
3. as we have already discussed, having a full working example in the dev server is something that helps me confidently propose this feature as an integration-tested reliable candidate for merging. In that sense, it is very much a part of this feature.

fabien0102:
@peterszerzo Perfect, I let you rebase and fix the `yarn.lock` and we should be able to merge 👍

TejasQ:
In response to [this](https://github.com/contiamo/restful-react/pull/72#issuecomment-432244767),

> it's more code to maintained

I think we have different thoughts behind how this will be used. [The way I see it](https://github.com/contiamo/restful-react/pull/72#discussion_r226651459), this code will _not_ be maintained because no part of it is public. Rather, I see it:

- being a quick demo of the functionality the PR describes,
- not changing with _all_ PRs, but just some specific ones,
- only being used in cases where we want to easily test and reproduce more real-world examples.

I think it's nice to have this as part of the "to test" list that goes into a PR for some cases that need more thorough integration testing, for the lifetime and purpose of validating their containing PR _only_.

I don't see much of a cost here for having this around. It's kind of like an [appendix](https://en.wikipedia.org/wiki/Appendix_(anatomy)): it doesn't do much, it _isn't included in the distributed package on npm_, it has no measurable cost to our users, but _can and does_ add value to our team as we develop features that might need more reassurance.

> If it's works in unit tests, it's works in the browser

This is a _very_ bold statement that I would carefully consider. Are you sure about this? There are enforcements that usually surprise us in browsers even today: [invalid/non-existent TLS](https://www.blog.google/products/chrome/milestone-chrome-security-marking-http-not-secure/), [HSTS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security), [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS), and [CORB](https://www.chromium.org/Home/chromium-security/corb-for-developers) that I'm not sure our unit tests would catch using `nock`. The dev-server allows us to test these cases in a variety of browsers ensuring that our PRs don't come with surprising behavior in real-world contexts.

TejasQ:
I would factor out that default resolve into another file, but besides that I don't see any major gotchas.

peterszerzo:
@TejasQ it's in `src/util` now.

fabien0102:
@peterszerzo Can your `git rebase -i` to avoid the 2/3 commits dealing with `packages-lock.json` and `yarn.lock`? :angel:
image

Issue #71, opened 2018-10-15
<Get /> may try to update state when component unmounts


carlesba:
**Describe the bug**

`fetch` method in `Get.tsx` could resolve the promise when component gets unmounted. This would call `this.setState` when component is not longer rendered what makes react to throw an error.

**To Reproduce**

In a test:

1. mount `Get` fetching for some resource.
2. create a `nock` instance that delays the response long enough to unmount the component before it resolves the request.
3. unmount the component and check the output. The error should be there.

**Expected behavior**

`setState` should not be called after component unmounts.

*A* solution here could be creating *cancelable promises* when fetching. Then keep track of pending requests in the instance of the component and cancelling the all the requests when component unmounts. `fetch` should not call `setState` when the request is cancelled.

**Screenshots**
<img width="1272" alt="unmounted" src="https://user-images.githubusercontent.com/1178459/46976211-1a1e5700-d0c9-11e8-994e-d6900b22d495.png">

TejasQ:
Great issue! Thanks for the report! Would you like to create a PR?

carlesba:
Sure! I'll try to put it in a PR. I'm a little bit busy those days so don't expect it soon.

TejasQ:
No compulsion whatsoever. 😄

Issue #12, opened 2018-07-16
Handle request concurrency


TejasQ:
From the comment in #11 ([specifically this](https://github.com/contiamo/restful-react/pull/11#discussion_r202736853)), we ought to come up with a way of handling multiple requests made in rapid succession.

fabien0102:
Even if it's not perfect yet, we can use the `debounce` prop to deal with this issue for now ;)

TejasQ:
We also handle this nicely in `Poll`, right?

TejasQ:
We also handle this nicely in `Poll`, right? @fabien0102

fabien0102:
Not yet, if you change the `path` props, nothing cancel the current request (I have deal with cancelling request on component unmount only for now ;) )

stereobooster:
With hooks we can handle it https://github.com/contiamo/restful-react/issues/101

TejasQ:
Closing because we cancel requests.

PR #74, opened 2018-10-20
Cancel pending request when unmounting

carlesba:
# Why

To avoid side effects from unmounted components.

# Related issue

#71

### DIFF
--- a/src/Poll.tsx
+++ b/src/Poll.tsx
@@ -4,6 +4,7 @@ import url from "url";
 
 import { InjectedProps, RestfulReactConsumer } from "./Context";
 import { GetProps, GetState, Meta as GetComponentMeta } from "./Get";
+import makeCancelable, { CancelablePromise } from "./util/makeCancelable";
 import { processResponse } from "./util/processResponse";
 
 /**
@@ -172,11 +173,8 @@ class ContextlessPoll<TData, TError> extends React.Component<
 
   private keepPolling = !this.props.lazy;
 
-  /**
-   * Abort controller to cancel the current fetch query
-   */
-  private abortController = new AbortController();
-  private signal = this.abortController.signal;
+  // List of requests to be cancelled when component unmounts
+  private pendingRequests: Array<CancelablePromise<any>> = [];
 
   private isModified = (response: Response, nextData: TData) => {
     if (response.status === 304) {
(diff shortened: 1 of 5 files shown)


carlesba:
Let me try implementing this using `AbortController`.

carlesba:
Let me try implementing this using just `AbortController`

carlesba:
Let me try this using `AbortController`

carlesba:
Let me try this using `AbortController`

src/Poll.tsx
```diff
@@ -172,11 +173,8 @@ class ContextlessPoll<TData, TError> extends React.Component<
 
   private keepPolling = !this.props.lazy;
 
-  /**
-   * Abort controller to cancel the current fetch query
-   */
-  private abortController = new AbortController();
-  private signal = this.abortController.signal;
+  // List of requests to be cancelled when component unmounts
+  private pendingRequests: Array<CancelablePromise<any>> = [];
```
fabien0102: Are you sure that is works as before? When I've checked the doc about how cancel a request, I found this `AbortController`, maybe cancelling the promise also works but I'm curious.

Also, I will prefer to have only one fetch running to avoid concurrencies issues (We don't want to have the result of the previous query if the second query is quicker). So ideally, if a previous request is running during a new fetch, you cancel the previous one :wink: This should also solve #12

src/util/makeCancelable.spec.ts
```diff
@@ -0,0 +1,42 @@
+import makeCancelable from "./makeCancelable";
+
+describe("makeCancelable", () => {
+  it("should resolve when promise resolves", async () => {
+    const future = new Promise(resolve => {
+      setTimeout(() => resolve("well done!"), 1000);
```
fabien0102: To have better tests (don't wait 1sc for each test) you can use this -> https://jestjs.io/docs/en/timer-mocks

fabien0102:
@carlesba Sorry, github is really broken today… at least you have some points for your next PR ;)

src/Poll.tsx
carlesba: Yeah! this works fine.

I tried this approach initially because I needed a way to check whether the request is still going before changing the state. Later I saw that you can also check that information from the signal, that's why I closed the PR.

carlesba:
@fabien0102 thanks for the tips! I'll try the `abortController` approach, I think it'll be much simpler than these.

PR #76, opened 2018-10-22
Avoid requests side effects once component unmounts

carlesba:
# Why

To avoid side effects from unmounted components.

# Related issue

#71


carlesba:
👍

carlesba:
I've just update the branch. I think that's all, what do you think?

fabien0102:
Rebased and tested on #89 
@carlesba Thanks for your amazing contribution on this issue 👍

PR #79, opened 2018-10-27
[POC] Hooks

fabien0102:
# Why

Just a little POC to try the new hooks react API https://reactjs.org/docs/hooks-intro.html

This PR is totally work in progress, it's just to open a discussion about what we can do or not with this new API.

### Usage
```tsx
import React from "react";
import { render } from "react-dom";

import { RestfulProvider, useGet } from "restful-react";

const ShowMeADog = () => {
  const { data, loading } = useGet<{ message: string }>({ path: "" });

  return loading ? <div>Loading…</div> : <img src={data ? data.message : ""} />;
};

const App = () => (
  <RestfulProvider base="https://dog.ceo/api/breeds/image/random">
    <ShowMeADog />
  </RestfulProvider>
);

render(<App />, document.getElementById("app"));
```

I tried with our `Get` component for now, and it's work pretty well. This avoid all ours nested components issues with the current pattern, and work with the `RestfulProvider` as before.

Due to the pattern, the `wait` is not possible (no render) and also we can't compose url as before (but it's really not blocker points from my perspective)

For now I have some problem with unit testing this new component, for a strange reason I only have the `Loading...` state and never the ``. I need to debug a bit more to know why, I also try to unit test the example of the doc, this one works perfectly, so it should be possible!

### How to play

1. Checkout the branch
1.  `yarn`
1. `yarn example`
1. Go to https://localhost:1234 and play with `/example/index.tsx`
1. Have fun 😃

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -92,6 +92,7 @@ typings/
 
 public
 lib
+.cache
 
 # Files managed by operational-scripts
 tslint.json
--- a/example/index.html
+++ b/example/index.html
@@ -1,0 +1,16 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <meta http-equiv="X-UA-Compatible" content="ie=edge">
+  <title>Restful react - hooks</title>
+</head>
+
+<body>
+  <div id="app"></div>
+  <script src="./index.tsx"></script>
+</body>
+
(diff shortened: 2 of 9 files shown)


micha-f:
@fabien0102 Wow, that was fast :)

TejasQ:
@fabien0102 I love this PR. Let's keep it around for reference when we have the time to move this over to hooks. For now, I will close it so it doesn't distract us from other PRs if that's okay.

Issue #78, opened 2018-10-26
Inconsistent render behavior when fetch rejects with net::ERR_SSL_PROTOCOL_ERROR


jgensler8:
**Describe the bug**

The fetch api can reject its returned promise if the server it is fetching data from returns an invalid certificate. While this behavior should result in the `<Get>` rendering it's `error` section, I instead see the progress indicator. While not 100% sure, I believe this is caused by the following line: https://github.com/contiamo/restful-react/blob/v5.2.1/src/Get.tsx#L202 as I figured there was some `await` call without a `catch`.

**To Reproduce**

Option 1:

If this ends up sharing correctly, this should be located here: https://codesandbox.io/s/9p3wmx7po

Option 2:

If using react-scripts, you should be able to switch some api call from `http://...` to `https://...`.

```
<RestfulProvider base="https://localhost:8080">
<Get path="/some/api/pets">
{(pets, { loading, error }) =>
    loading ? (
       Loading...
    ) : (
        <div>
        You should only see this after things are loaded.
        {error ? (
            <div>"OH NO!"</div>
        ) : (
            <div>success!</div>
        )}
        </div>
)}
</Get>
</RestfulProvider>
```

Option 3:

1. Create a server that returns a different certificate than the domain used to query it (Subject Alternative Name or Common Name is different)
2. Send a GET request to this server
3. observe UI behavior

**Expected behavior**

If the GET request fails because in invalid certificates, the `error` section should render.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: OSX Darwin Kernel Version 15.6.0
 - Browser Chrome
 - Version `v5.2.1`

**Additional context**

There might be other edge cases (maybe CORS or timeouts?) that cause this same behavior though I don't have any evidence of that.

TejasQ:
Thanks for the report! Would you like to pair up on a PR? 😄

jgensler8:
I should be able to find some time this week to work on a PR.

I started by trying to run the test suite (`npm test`). I had a few errors and I want to make sure that I can easily test the changes on my laptop before developing.

First, I got the following:

```
 FAIL  src/Poll.test.tsx
  ● Test suite failed to run

    TypeScript diagnostics (customize using `[jest-config].globals.ts-jest.diagnostics` option):
    src/Poll.test.tsx:419:23 - error TS7006: Parameter 'data' implicitly has an 'any' type.

    419       const resolve = data => ({ ...data, too: "bar" });
                              ~~~~
[...]

PR #84, opened 2018-10-30
Define behavior for fetch failure for Get, Mutate, and Poll.

jgensler8:
Get and Mutate are pretty clear at how they should handle failure. Poll's behavior is to completely stop when a fetch fails. I don't understand how this is useful as you are probably using Poll to reach a desire state with your 'until' and can deal with the types of errors there. I haven't exposed the fetch errors with state because this would likely result in the until's function signature changing and I don't want to break API versions.

# Why

relates to #78

### DIFF
--- a/src/Mutate.tsx
+++ b/src/Mutate.tsx
@@ -138,25 +138,35 @@ class ContextlessMutate<TData, TError> extends React.Component<
       },
     });
 
-    const response = await fetch(request);
-    const { data, responseError } = await processResponse(response);
+    return fetch(request)
+      .catch(error => {
+        throw {
+          message: `Failed to fetch: ${error}`,
+          data: "",
+        };
+      })
+      .then(response => processResponse(response))
+      .then(processedResponse => {
+        const { response, data, responseError } = processedResponse;
 
-    if (!response.ok || responseError) {
-      const error = { data, message: `Failed to fetch: ${response.status} ${response.statusText}` };
+        if (!response.ok || responseError) {
+          throw { data, message: `Failed to fetch: ${response.status} ${response.statusText}` };
+        }
 
-      this.setState({
-        loading: false,
+        this.setState({ loading: false });
+        return data;
+      })
+      .catch(error => {
+        this.setState({
+          loading: false,
+        });
+
+        if (!this.props.localErrorOnly && this.props.onError) {
+          this.props.onError(error, () => this.mutate(body, mutateRequestOptions));
+        }
+
+        throw error;
       });
-
-      if (!this.props.localErrorOnly && this.props.onError) {
-        this.props.onError(error, () => this.mutate(body, mutateRequestOptions));
-      }
-
-      throw error;
-    }
-
-    this.setState({ loading: false });
-    return data;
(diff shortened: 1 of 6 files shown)


src/Mutate.tsx
```diff
@@ -138,25 +138,35 @@ class ContextlessMutate<TData, TError> extends React.Component<
...
-      this.setState({
-        loading: false,
+    return fetch(request)
+      .catch(error => {
+        throw {
+          message: `Failed to fetch: ${error}`,
+          data: "",
+        };
+      })
+      .then(response => processResponse(response))
```
fabien0102:
```suggestion
      .then(processResponse)
```

src/Poll.tsx
```diff
@@ -222,42 +222,46 @@ class ContextlessPoll<TData, TError> extends React.Component<
...
+            loading: false,
+            lastResponse: response,
+            previousData: prevState.data,
+            data,
+            error: null,
+            lastPollIndex: response.headers.get("x-polling-index") || undefined,
+          }));
+        }
+      })
+      .catch(error => {
```
fabien0102: mmm, if I'm following well, in case of a cancelled request, we are doing `.catch(() => throw)` -> `.catch(this.props.onError)`, I'm wonder if we can test this in our test suite, maybe with an early `cleanup` or something like this (just to sure that we don't have any memory leak in this part)

src/Poll.tsx
```diff
@@ -222,42 +222,46 @@ class ContextlessPoll<TData, TError> extends React.Component<
...
         };
-        this.setState({ loading: false, lastResponse: response, error });
+      })
+      .then(response => processResponse(response))
+      .then(processedResponse => {
+        const { response, data, responseError } = processedResponse;
+
+        if (!this.keepPolling) {
+          // Early return if we have stopped polling to avoid memory leaks
+          return;
```
fabien0102: Carreful, an early return doesn't work in a `.then()`

```js
Promise.resolve("hello")
  .then(() => {
    return 
  })
  .then(() => console.log("plop"))

// I have "plop" in my console
```

TejasQ:
[Inconsistent render behavior when fetch rejects with net::ERR_SSL_PROTOCOL_ERROR](https://trello.com/c/4NhNu50G/549-inconsistent-render-behavior-when-fetch-rejects-with-neterrsslprotocolerror)

TejasQ:
@jgensler8 still interested in looking at this together with us?

jgensler8:
I won't be able to get to this PR until the end of this week. If this issue is a blocker for others, feel free to fix and deny this PR.

TejasQ:
No worries. 🙂  We can pick it up later. I was just interested in the status in order for us to all be on the same page.

TejasQ:
@jgensler8 any updates here? I understand things get busy sometimes. There is no problem at all, but I'd rather close this if you're unable to find the time. 🙂

jgensler8:
Yeah let's close this for now and I can reopen if I get closer to a fix.
The issue is still open so at least some of the work is documented.

PR #87, opened 2018-10-31
Fix an infinite loop on fetching without a resolve function

TejasQ:
# Why

On 6.0.0, we accidentally introduced a regression that rerenders the component infinitely if a `Get` does not have a `resolve` function and the request happens to error. This was introduced by incorrectly comparing resolve functions.

As a way of ensuring function-referential-equivalence, we now stringify the functions to make sure we only rerender when they _actually_ change.

### DIFF
--- a/src/Get.test.tsx
+++ b/src/Get.test.tsx
@@ -567,23 +567,14 @@ describe("Get", () => {
         .persist();
       const children = jest.fn();
       children.mockReturnValue(<div />);
-      const resolve = a => a;
-      /**
-       * A new instance of RestfulProvider is created on every rerender.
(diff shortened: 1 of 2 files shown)


TejasQ:
@peterszerzo your concerns are really valid and I think you make excellent points. I opened and expedited the merging of this PR in order to _fix the project quickly_ in order to prevent new `npm install restful-react`s resulting in a broken and disappointing experience to adopters.

I think the `.toString()`ing of the functions is not fully ideal, especially because it comes with ambiguous intent. I would love to discuss this more, perhaps in an issue, and solve it in a much more elegant way, including even rethinking the "feature" in the first place.

PR #88, opened 2018-10-31
Fix base path propagation to ignore parentPath

TejasQ:
# Why

Currently, the following snippet behaves a little unpredictably. Consider,

```jsx
        <RestfulProvider base="https://my-awesome-api.fake/eaegae">
          <Get path="/LOL">
            {() => (
              <Mutate verb="POST" base="https://my-other-api.fake" path="">
                {children}
              </Mutate>
            )}
          </Get>
        </RestfulProvider>
```

this makes a request to `https://my-other-api.fake/LOL` when the mutate function is called. This is incorrect. It needs to instead, ignore any parent path and defer to `Mutate`'s `path` property (the empty string) _only_, without composing the parent `Get`'s `path` property.

In order for this function to function (pun intended) correctly, the snippet _should_ send a request to `https://my-other-api.fake/`.

This PR fixes this and adds a test case across both `Get` _and_ `Mutate`.

### DIFF
--- a/src/Get.tsx
+++ b/src/Get.tsx
@@ -60,6 +60,11 @@ export interface GetProps<TData, TError> {
    */
   path: string;
   /**
+   * @private This is an internal implementation detail in restful-react, not meant to be used externally.
+   * This helps restful-react correctly override `path`s when a new `base` property is provided.
+   */
+  __internal_hasExplicitBase?: boolean;
+  /**
    * A function that recieves the returned, resolved
    * data.
    *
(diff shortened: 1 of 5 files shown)


src/Get.tsx
```diff
@@ -277,7 +287,7 @@ function Get<TData = any, TError = any>(props: GetProps<TData, TError>) {
     <RestfulReactConsumer>
       {contextProps => (
         <RestfulReactProvider {...contextProps} parentPath={composePath(contextProps.parentPath, props.path)}>
-          <ContextlessGet {...contextProps} {...props} />
+          <ContextlessGet {...contextProps} {...props} __internal_hasExplicitBase={Boolean(props.base)} />
```
peterszerzo: Is this introduced to keep track of what was initially passed in, anticipating an override by `defaultProps`? If so, I think it'd be clearer to remove `defaultProps` and implement defaulting logic manually inside.

TejasQ: > Is __internal_hasExplicitBase introduced to keep track of what base prop was initially passed in, anticipating an override by defaultProps?

The motivation behind this prop was described in the [type definition](https://github.com/contiamo/restful-react/pull/88/files#diff-63d62a288bee27bda4e83b106d69c46dR62). 🙂 

Essentially, [_everything_ has a `base` prop implicitly via context](https://github.com/contiamo/restful-react/blob/master/src/Get.tsx#L279), and paths compose as you'd expect. 

💩 really hits the fan when a `base` is passed explicitly because the `parentPath` is still appended to the `base`, as outlined in the description of this PR. This is unexpected. It certainly was unexpected in Labs. 

We need some way of knowing "hey, I have an explicit base reset" in order to [tell restful-react that it should disregard `parentPath`](https://github.com/contiamo/restful-react/blob/fix/base-propagation/src/Get.tsx#L223-L227), and so this _internal, non-public-API prop_ was introduced.

peterszerzo: Sweet, it makes sense now, thanks for clearing it up!

PR #90, opened 2018-11-05
Add test to verify polling doesn't happen if no index

TejasQ:
# Why
To correctly implement the polling spec, we should not continue polling if we do not have a `x-polling-index` in the response.

This seems to already be the case, but I've added an extra test case just in case.


TejasQ:
[Stop polling if don't have x-polling-index in response](https://trello.com/c/Fd140Huc/537-stop-polling-if-dont-have-x-polling-index-in-response)

PR #94, opened 2018-11-13
Add Response in the onError callback

fabien0102:
# Why

The idea is to add `response` as extra argument of `onError` to have access to the `url`, `status`, etc for better debugging


fabien0102:
[Add failed requests to the debug stack](https://trello.com/c/RZahdwN0/589-add-failed-requests-to-the-debug-stack)

fabien0102:
I was a bit optimistic about testing… let me fix this ;)

PR #95, opened 2018-11-20
Fix React import on generated d.ts

diegochavez:
React should be imported with * because of React.ReactNode is not being exported with React Object I'm using TSX

# Why
After installing the package it throws an error React `has no default export`
image

### DIFF
--- a/src/Poll.tsx
+++ b/src/Poll.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import * as React from "react";
 import equal from "react-fast-compare";
 
 import { InjectedProps, RestfulReactConsumer } from "./Context";


fabien0102:
First of all, good catch, `Poll` is actually the only file without the `* as` pattern 😮 

But it's really strange, we are using this project in many of our products and never had this error before and `allowSyntheticDefaultImports: true` in the `tsconfig.json` should solve this 🤔

TejasQ:
WOW @diegochavez, great work! Thanks for this amazing PR! 🔥

TejasQ:
Published to Restful React v6.1.1!

diegochavez:
Awesome thanks, great library guys! :+1:

TejasQ:
💞

PR #100, opened 2019-02-01
Fix infinite request loop because of queryParam equality

TejasQ:
# Why
We were comparing queryParams for equality and it was always false, leading to infinite rerenders and DDoS attacks.

This PR fixes this.


LucasRoesler:
[Infinity loop of 404 when the configmap is wrong](https://trello.com/c/5useAMy1/135-infinity-loop-of-404-when-the-configmap-is-wrong)

PR #103, opened 2019-03-11
useGet hooks

fabien0102:
# Why

Begin the migration to hooks to enable super powers 💯 

This PR introduce all the feature of `Get` but in hook implementation under `useGet`.


src/Context.tsx
```diff
@@ -55,9 +55,11 @@ export default class RestfulReactProvider<T> extends React.Component<RestfulReac
         }}
       >
         {children}
-      </Provider>
+      </Context.Provider>
     );
   }
 }
 
+const RestfulReactConsumer = Context.Consumer;
```
stereobooster:
```suggestion
export const RestfulReactConsumer = Context.Consumer;
```

src/useGet.tsx
```diff
@@ -0,0 +1,182 @@
...
+  const prevProps = useRef<UseGetProps<TData, TQueryParams>>();
+  useEffect(() => {
+    const { base, path, resolve, queryParams } = prevProps.current || {
+      base: null,
+      path: null,
+      resolve: null,
+      queryParams: null,
+    };
+
+    if (
```
stereobooster: Consider `useDeepCompareEffect` https://gist.github.com/kentcdodds/fb8540a05c43faf636dd68647747b074

src/useGet.tsx
```diff
@@ -0,0 +1,182 @@
...
+  }, [props.path, props.base, props.resolve, props.queryParams]);
+
+  return {
+    ...state,
+    absolutePath: url.resolve(
+      props.base || context.base,
+      props.queryParams ? `${props.path}?${qs.stringify(props.queryParams)}` : props.path,
+    ),
+    refetch: (options: Partial<Omit<UseGetProps<TData, TQueryParams>, "lazy">> = {}) =>
+      fetchData({ ...props, ...options }, state, setState, context, abortControllers.current),
```
stereobooster: `state` can get stale here. Suggestion is to use `useReducer` see https://twitter.com/kentcdodds/status/1102647400915582981 or https://codesandbox.io/s/71r1rqnvr6

fabien0102: Since this function is directly render, I don't see how 😅

src/useGet.tsx
fabien0102: It's exactly the same pattern 😅 I think adding an abstraction here will complexify the code a bit more in this case.

stereobooster: it will simplify `useGet`, by removing non essential code from `useGet`.

src/useGet.tsx
```diff
@@ -0,0 +1,182 @@
...
+  }
+}
+
+export function useGet<TData = any, TError = any, TQueryParams = { [key: string]: any }>(
+  props: UseGetProps<TData, TQueryParams>,
+) {
+  const context = useContext(Context);
+
+  const fetchData = useCallback<typeof _fetchData>(
+    typeof props.debounce === "object"
```
TejasQ: Can we please refactor this to not be a nested ternary expression? 🙏

src/useGet.tsx
```diff
@@ -0,0 +1,182 @@
...
+    data: null,
+    response: null,
+    loading: !props.lazy,
+    error: null,
+  });
+
+  const abortControllers = useRef([new AbortController()]);
+
+  const prevProps = useRef<UseGetProps<TData, TQueryParams>>();
+  useEffect(() => {
```
TejasQ: Please add a new line here to separate this from `prevProps`. I thought this was still the implementation of `prevProps` and had to scroll up/look twice.

```suggestion

  useEffect(() => {
```

src/useGet.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+  const [state, setState] = useState<GetState<TData, TError>>({
+    data: null,
+    response: null,
+    loading: !props.lazy,
+    error: null,
+  });
+
+  const abortControllers = useRef([new AbortController()]);
+
```
TejasQ: Let's group `const`s together to better separate them from effects.

```suggestion
```

src/useGet.tsx
TejasQ: I agree with @stereobooster, but this isn't a showstopper, of course.

src/useGet.tsx
TejasQ: Let's deal with the issue of stale state if:

- We can clearly produce a failing test/reproduce it.
- It is an _actual_ issue in one of our projects instead of a future, conceptual issue that _may not exist_.

If the claim of stale state can be substantiated, let's substantiate it and fix it. If not, let's merge this now.

src/useGet.tsx
fabien0102: It's perfect! (just kidding ^^ but it's my best option here ;) )

src/util/useDeepCompareEffect.ts
```diff
@@ -0,0 +1,35 @@
+import isEqualWith from "lodash/isEqualWith";
+import React, { useEffect, useRef } from "react";
+
+function useDeepCompareMemoize(value: Readonly<any>) {
+  const ref = useRef<any>();
+
+  if (
+    !isEqualWith(value, ref.current, (a, b) => {
```
stereobooster: can we factor out this as a separate function, please?

src/useGet.tsx
```diff
@@ -0,0 +1,161 @@
...
+    });
+  }
+}
+
+export function useGet<TData = any, TError = any, TQueryParams = { [key: string]: any }>(
+  props: UseGetProps<TData, TQueryParams>,
+) {
+  const context = useContext(Context);
+
+  const fetchData = useCallback<typeof _fetchData>(
```
stereobooster: Debounce can be called after component will be unmounted. We need something like `useDebounceCallback` https://github.com/xnimorz/use-debounce/blob/master/src/callback.js

src/useGet.tsx
```diff
@@ -0,0 +1,161 @@
...
+      const error = {
+        message: `Failed to fetch: ${response.status} ${response.statusText}${responseError ? " - " + data : ""}`,
+        data,
+        status: response.status,
+      };
+
+      setState({ ...state, loading: false, error });
+
+      if (!props.localErrorOnly && context.onError) {
+        context.onError(error, () => _fetchData(props, state, setState, context, abortController), response);
```
stereobooster: I guess, state can get stale here, because it is passed through context

src/useGet.tsx
TejasQ: Agree with @stereobooster – let's use this and then ship. 🚢

src/useGet.tsx
TejasQ: If we say something _can_ happen, let's substantiate the claim with a failing test case or a concrete example (maybe codesandbox?). Else, let's ship.

src/useGet.tsx
```diff
@@ -0,0 +1,171 @@
...
+    setState({ ...state, error: null, loading: true });
+  }
+
+  const requestOptions =
+    (typeof props.requestOptions === "function" ? props.requestOptions() : props.requestOptions) || {};
+  requestOptions.headers = new Headers(requestOptions.headers);
+
+  const request = new Request(
+    url.resolve(base, queryParams ? `${path}?${qs.stringify(queryParams)}` : path),
+    requestOptions,
```
TejasQ:
```suggestion
    { ...requestOptions, signal }
```

PR #113, opened 2019-04-07
Add support for requestOptions as a method to Poll

### LINKED ISSUE #112: Poll doesn't support function for requestOptions same way as Get does
**Describe the bug**
Poll doesn't support function for requestOptions same way as Get does

**To Reproduce**
1. Create a RestfulProvider with requestOptions as a function returning some header
2. Use Get and Poll inside
3. Open devtools and see that headers on poll requests are not sent

**Expected behavior**
Behaviour should be the same for Poll and Get

**Additional context**
I found this issue while using a version of safari that requred `credentials: 'same-origin'` in request options and it took me a while to figure out

vxsx:
# Why

Fixes #112 

I'm not sure this is a proper way, but it works. Get / Poll seem to be quite different and sadly I didn't have much time to dig into the source code.

### DIFF
--- a/src/Poll.tsx
+++ b/src/Poll.tsx
@@ -1,3 +1,4 @@
+import merge from "lodash/merge";
 import * as qs from "qs";
 import * as React from "react";
 import equal from "react-fast-compare";
@@ -351,10 +352,7 @@ function Poll<TData = any, TError = any, TQueryParams = { [key: string]: any }>(
           <ContextlessPoll
             {...contextProps}
             {...props}
-            requestOptions={{
-              ...contextRequestOptions,
-              ...propsRequestOptions,
-            }}
+            requestOptions={merge(contextRequestOptions, propsRequestOptions)}
           />
         );
       }}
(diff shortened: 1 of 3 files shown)


src/Poll.tsx
```diff
@@ -339,16 +339,25 @@ function Poll<TData = any, TError = any, TQueryParams = { [key: string]: any }>(
...
+            : contextProps.requestOptions || {};
+        const propsRequestOptions =
+          typeof props.requestOptions === "function" ? props.requestOptions() : props.requestOptions || {};
+
+        return (
+          <ContextlessPoll
+            {...contextProps}
+            {...props}
+            requestOptions={{
+              ...contextRequestOptions,
```
fabien0102: Almost perfect! But you need to use `merge` from lodash here, so we can add a header without erase these set in the Provider

Example:
https://github.com/contiamo/restful-react/blob/master/src/useGet.tsx#L87

and the associate test:
https://github.com/contiamo/restful-react/blob/eaa37d294e7173a7a1de6a75ac832ab85283a112/src/useGet.test.tsx#L512-L533

vxsx: Right, good catch, thanks!

Fixed and added a test.

TejasQ:
@vxsx, you're a legend. Thanks for your contribution! 🎉

PR #114, opened 2019-04-16
Fix useGet url composition pattern

fabien0102:
# Why

By lazyness, I've choose `url.resolve` composition pattern when I implement the first version of `useGet`, the problem is that it required to add a trailing slash to the `Provider.base` to be able to use the library…

To resolve this, I update a bit the pattern, to permit more convenient url composition, not sure that is yet perfect but it will be test in our internals applications to validate the usage and pattern :wink:

### DIFF
--- a/src/useGet.tsx
+++ b/src/useGet.tsx
@@ -57,10 +57,10 @@ export interface UseGetProps<TData, TQueryParams> {
 }
 
 function resolvePath<TQueryParams>(base: string, path: string, queryParams: TQueryParams) {
-  const escapedBase = base.endsWith("/") ? base : `${base}/`;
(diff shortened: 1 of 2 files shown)


src/useGet.tsx
```diff
@@ -56,6 +56,13 @@ export interface UseGetProps<TData, TQueryParams> {
     | number;
 }
 
+function resolvePath<TQueryParams>(base: string, path: string, queryParams: TQueryParams) {
+  const escapedBase = base.endsWith("/") ? base : `${base}/`;
+  const escapedPath = path.startsWith("/") ? path.slice(1) : path;
```
mpotomin: `escaped` confuses me here - makes me think about url encoding, how about `appendedBase` and `trimmedPath`?

PR #126, opened 2019-05-17
Generate `useMutate` components

fabien0102:
# Why

All the power of this library is the open-api generation, this permit to have auto-completion of our backend api directly in react.
I have the honor to introduce `useMutate` part of this auto-generation! now every route can be generate and use with hooks :tada:

## Example

For the following spec:

```yaml
post:
      description: Creates a new pet in the store.  Duplicates are allowed
      operationId: addPet
      requestBody:
        description: Pet to add to the store
        required: true
        content:
          application/json:
            schema:
              $ref: "#/components/schemas/NewPet"
      responses:
        "200":
          description: pet response
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Pet"
        default:
          description: unexpected error
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Error"
```

I will have the following components generated:

```tsx
export type AddPetProps = Omit<MutateProps<Pet, Error, void, NewPet>, \\"path\\" | \\"verb\\">;


export const AddPet = (props: AddPetProps) => (
  <Mutate<Pet, Error, void, NewPet>
    verb=\\"POST\\"
    path={\`/pets\`}
    {...props}
  />
);

export type UseAddPetProps = Omit<UseMutateProps<Pet, void>, \\"path\\" | \\"verb\\">;


export const useAddPet = (props: UseAddPetProps) => useMutate<Pet, Error, void, NewPet>(\\"POST\\", \`/pets\`, props);

```

So I can do in my application:

```tsx
import { useAddPet } from "./petstore-generated";

const App = () => {
  const {mutate: addPet, loading, error} = useAddPet();

  return (<div>
     {loading && <div>Loading…</div>}
     {error && <div>{error.message}</div>}
    <button onClick={() => addPet({name: "Medor"})}>Add a pet</button>
  </div>)
};
```

This is of course fully typesafe, and you have autocompletion on `addPet({name: "Medor"})` 💯 

## Validator

I also use this branch to continue on our goal to improve the quality and strictness of our specifications.

Now, https://github.com/IBM/openapi-validator is integrate by default, and we can just add a `.validaterc` in any project to configure the openapi linter 😈 

Note: This validator is really strict! I've added a `--no-validation` flag to be able to have a more smooth migration to this dream 😅


src/Mutate.tsx
```diff
@@ -16,7 +16,10 @@ export interface States<TData, TError> {
   error?: GetState<TData, TError>["error"];
 }
 
-export type MutateMethod<TData, TRequestBody> = (data?: TRequestBody) => Promise<TData>;
+export type MutateMethod<TData, TRequestBody> = (
+  data: TRequestBody,
```
fabien0102: `data` is not anymore optional, so we can define `string` for DELETE operation.
This is to have a bit more type safety, `mutate()` is not possible anymore if a body is required.

src/scripts/ibm-openapi-validator.d.ts
```diff
@@ -0,0 +1,21 @@
+declare module "ibm-openapi-validator" {
```
TejasQ: Sad that it's not TS.

TejasQ: Can we contribute this to @DefinitelyTyped?

src/scripts/import-open-api.ts
```diff
@@ -468,24 +470,73 @@ export interface ${pascal(name)}Response ${type}`;
...
 };
 
+/**
+ * Validate the spec with ibm-openapi-validator (with a custom pretty logger).
+ *
+ * @param schema openAPI spec
+ */
+const validate = async (schema: OpenAPIObject) => {
+  // tslint:disable:no-console
+  const log = console.log;
```
TejasQ: Maybe export this from [that other place you do it](https://github.com/contiamo/restful-react/pull/126/files#diff-2914e40119ca01b2d64663f0a82bb278R10)?

src/scripts/import-open-api.ts
```diff
@@ -468,24 +470,73 @@ export interface ${pascal(name)}Response ${type}`;
...
+    log(chalk.red("(!) Errors"));
+    errors.forEach(i =>
+      log(
+        chalk.red(`
+Message : ${i.message}
+Path    : ${i.path}`),
+      ),
+    );
+  }
+  // tslint:enable:no-console
```
TejasQ: ?

fabien0102: I've disabled the `no-console` rule for the block, so I re-enable it ^^

src/scripts/import-open-api.ts
fabien0102: It's more have only one `no-console`, I prefer to not export this hack ^^ So you need to ask yourself "is it a real usecase or just a console.log for debugging?"

src/scripts/ibm-openapi-validator.d.ts
fabien0102: 24 stars on github, I've tried but without a lot of hope 😅 I will contribute to definitelyTyped to add this later.

src/scripts/import-open-api.ts
fabien0102: image

Issue #136, opened 2019-07-05
Avoid Error Updates on Unmounted Component


AJHenry:
**Describe the bug**
The `Get` component has an abort handler that will cancel any state updates after the component umounts. However, since the fetch call is wrapped in a try catch block, it will bypass the abort check and update state resulting in the classic `Warning: Can't perform a React state update on an unmounted component`

This is because the abort check is after the catch block intercepts the error thrown by `fetch`

https://github.com/contiamo/restful-react/blob/529bdb35c9c6baec36cd34ccd76510726b142332/src/Get.tsx#L256-L262

I was able to fix this issue by putting the abort check in the catch block as well

https://github.com/AJHenry/restful-react/blob/8627e00f58cfc8c69d050c0d9a811717aa6e0533/src/Get.tsx#L381-L385

**To Reproduce**
1. Mount the Get component
2. Call an API that will result in an error such as 401, 404, etc
3. Unmount the component before the call finishes and the error is thrown

**Expected behavior**
The state should not be updated when the component is unmounted, regardless of whether there is an error or not

**Additional context**
I would open a PR for this issue myself but I am unable to reliably reproduce the issue in a unit test (as `setState` on unmounted components are fairly hard to test)

TejasQ:
Thanks for the issue, @AJHenry! I'd suggest opening a PR regardless and we can collaborate towards a good working test together. What do you think?

AJHenry:
@TejasQ Sounds good, I can open a PR later today

stereobooster:
@fabien0102 @AJHenry is this resolved issue? Can we close the ticket?

AJHenry:
@stereobooster this should be fixed as #137 was merged in awhile back

PR #137, opened 2019-07-08
Add fix for #136

AJHenry:
# Why

This PR is to provide a fix for #136 where the fetch request could throw an error after the component was unmounted

## Note

I have not yet been able to create a unit test that consistently reproduces this error as this is a race condition with an error

### DIFF
--- a/src/Get.tsx
+++ b/src/Get.tsx
@@ -285,6 +285,12 @@ class ContextlessGet<TData, TError, TQueryParams> extends React.Component<
       this.setState({ loading: false, data: resolved.data, error: resolved.error });
       return data;
     } catch (e) {
+      // avoid state updates when component has been unmounted
+      // and when fetch/processResponse threw an error
+      if (this.signal.aborted) {
+        return;
+      }
+
       this.setState({
         loading: false,
         error: {


TejasQ:
What if we hoist the check for the abort signal above the try/catch block and simply return? It doesn't really make sense to call setState if a request is aborted anyway.

That way we have the check only once.

AJHenry:
I think that would be a better solution, however what if the component unmounts during the fetch request. I am not too familiar with the AbortController but if we passed the abort check, called fetch, then the component unmounted, program execution would continue and we would have to perform the check afterwards anyway?

TejasQ:
[The component cancels on unmount](https://github.com/contiamo/restful-react/blob/529bdb35c9c6baec36cd34ccd76510726b142332/src/Get.tsx#L203) so it can't unmount _during_ the fetch request.

AJHenry:
Right, so when the component will unmount, it will set the abort flag in the AbortController.

So if a user visiting a page with this component on it, the component will mount, and call the fetch request. Now say this fetch request takes 5 seconds to load (because of a slow API), the user can click away from the page, thereby unmounting the component in the middle of the fetch request. 

Now the AbortController will have it's abort flag set and cancel the fetch request, then [the fetch() promise rejects with an AbortError](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) thereby being caught in the try-catch again

fabien0102:
@AJHenry Published on version `8.1.0` 😉

PR #141, opened 2019-07-29
Add a onMutate callback.

fabien0102:
# Why

I need a onMutate callback to unvalidate some cache, the plan is to slowly introduce some caching strategy, not yet sure how we want to this but a `onMutate` callback can be useful for many situation.

For now I just have a `memoize` function around our restful-react getter, and the idea is to unvalidate the cache on every related mutation 😉

### DIFF
--- a/src/useMutate.test.tsx
+++ b/src/useMutate.test.tsx
@@ -218,6 +218,20 @@ describe("useMutate", () => {
       }
     });
 
+    it("should call onMutation", async () => {
+      nock("https://my-awesome-api.fake")
+        .post("/")
+        .reply(200, { ok: true });
+
+      const wrapper = ({ children }) => (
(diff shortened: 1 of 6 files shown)


src/useMutate.test.tsx
```diff
@@ -407,7 +421,7 @@ describe("useMutate", () => {
       }
 
       type UseDeleteMyCustomEndpoint = Omit<
-        UseMutateProps<MyCustomEnpointResponse, MyCustomEnpointQueryParams>,
+        UseMutateProps<MyCustomEnpointResponse, MyCustomEnpointQueryParams, {}>,
```
fabien0102: I'm so stupid… Fixing the test but not the generator… 🤦‍♂

PR #144, opened 2019-08-01
[WIP] New GitHub auth for open-api generator

fabien0102:
# Why

We was using the classic dev token, store in node_module, it's a bit hidded in github and can be sometime confusing/repetitive.

This PR implement a real github auth flow, with a webpage asking for authorization. No token are stored locally and the user need to approve the application the first time only.

The flow is the following:
1. `restful-react import --github contiamo:restful-react:master:my-spec.yaml`
1. open the browser to ask authorization (github website) (only the first time)
1. import your spec

# Todo
- [ ] Set up the correct lambda url (need to be deployed on master to test the entire workflow)
- [ ] Remove the `--github-auth-url` option (test purpose only)


fabien0102:
My function is not deployed… I need to fix this, I will do this later…

fabien0102:
I don't have time for this, and I'm not sure it's really better to use…

PR #157, opened 2019-10-22
Support Global queryparams

TejasQ:
# Why
In some cases, we'd like query params to be inherited by all hooks/components under a given provider. This PR makes this possible.

### DIFF
--- a/src/Context.tsx
+++ b/src/Context.tsx
@@ -27,6 +27,11 @@ export interface RestfulReactProviderProps<T = any> {
    * to deal with your retry locally instead of in the provider scope.
    */
   onError?: (err: any, retry: () => Promise<T | null>, response?: Response) => void;
+  /**
+   * Any global level query params?
+   * **Warning:** it's probably not a good idea to put API keys here. Consider headers instead.
+   */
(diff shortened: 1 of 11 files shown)


src/Context.tsx
```diff
@@ -27,6 +27,11 @@ export interface RestfulReactProviderProps<T = any> {
    * to deal with your retry locally instead of in the provider scope.
    */
   onError?: (err: any, retry: () => Promise<T | null>, response?: Response) => void;
+  /**
+   * Any global level query params?
+   * **Warning:** it's probably not a good idea to put API keys here. Consider headers instead.
```
mpotomin: ❤️

src/Get.tsx
```diff
@@ -245,7 +246,9 @@ class ContextlessGet<TData, TError, TQueryParams> extends React.Component<
       } else {
         url = composeUrl(base!, parentPath!, requestPath || path || "");
       }
-      if (this.props.queryParams) {
+
+      // We use ! because it's in defaultProps
```
mpotomin: 👍

PR #169, opened 2019-11-06
Support pulling spec from URL

qw-in:
# Why

We are using [fastapi](https://github.com/tiangolo/fastapi) which generates the openapi spec so it is more convenient just to pull it right off the server

This pr adds (naive) support for grabbing specs from any http endpoint via the `-url or -u` flag

PS - I could not get tslint to run...kept throwing

```
<snip>/restful-react/src/scripts/tests' is not included in project.
```


.gitignore
```diff
@@ -105,4 +105,5 @@ examples/*.tsx
 
 # Files managed by operational-scripts
 tslint.json
+tsconfig.json
```
qw-in: This got auto-generated, happy to remove if not wanted

TejasQ:
Hey @qw-in! Amazing work! We'll take a look at it in a little more detail soon and respond.

.gitignore
fabien0102: Please remove it… this is due to a bad behaviour of `@operational/script` (a false good idea 😅) cc @TejasQ

fabien0102:
Expect the little fix in `.gitignore` this looks good to me 👌

PR #174, opened 2019-11-14
Fix type definition of refetch()

ysgk:
# Why

`refetch()` type definition is outdated. Actually, it receives options as its first argument.

https://github.com/contiamo/restful-react/blob/59ccaa2a487ca8844c86a1be7aca65cc7949dcc1/src/useGet.tsx#L221


TejasQ:
OH EM GEE! 🙀 

Thank you for pointing this out! Let's fix it! Would you like to contribute by making a PR?

TejasQ:
Thank you so much for your amazing contribution, @ysgk! 🚀

Issue #156, opened 2019-10-22
Support `discriminator` in openapi generator


fabien0102:
**Is your feature request related to a problem? Please describe.**

We actually don't deal with the `discrimator` pattern, the actual workaround is to use `oneOf` combine with `enum` like this:

```yaml
Error:
  oneOf:
  - "$ref": "#/components/schemas/GeneralError"
  - "$ref": "#/components/schemas/FieldError"
GeneralError:
  type: object
  required:
  - type
  - message
  properties:
    type:
      type: string
      enum:
      - GeneralError
    message:
      type: string
FieldError:
  type: object
  required:
  - type
  - message
  - key
  properties:
    type:
      type: string
      enum:
      - FieldError
    message:
      type: string
    key:
      type: string
```

**Describe the solution you'd like**
The goal is to support the discrimator pattern as describe here -> https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/

The previous example will become:

```yaml
Error:
  oneOf:
  - "$ref": "#/components/schemas/GeneralError"
  - "$ref": "#/components/schemas/FieldError"
  discriminator:
    propertyName: type
    mapping:
      GeneralError: "#/components/schemas/GeneralError"
      FieldError: "#/components/schemas/FieldError"
GeneralError:
  type: object
  required:
  - type
  - message
  properties:
    type:
      type: string
    message:
      type: string
FieldError:
  type: object
  required:
  - type
  - message
  - key
  properties:
    type:
      type: string
    message:
      type: string
    key:
      type: string

```

The generated types for the two previous examples should be the same:

```ts
export interface GeneralError {
  type: "GeneralError";
  message: string;
}

export interface FieldError {
  type: "FieldError";
  message: string;
  key: string;
}

export type Error = GeneralError | FieldError;
```

PR #175, opened 2019-11-14
Deal with discriminator pattern

fabien0102:
Deal with the discriminator pattern to solve #156 

This is definitely not a full specs implementation, we only resolve discriminator inside `components/schemas`.

We can of course iterate later when we/somebody have more advanced needs.

### DIFF
--- a/src/scripts/tests/__snapshots__/import-open-api.test.ts.snap
+++ b/src/scripts/tests/__snapshots__/import-open-api.test.ts.snap
@@ -12,6 +12,12 @@ export type Pet = NewPet & {id: number};
 
 export interface NewPet {name: string; tag?: string}
 
+export type CatOrDog = Cat | Dog;
+
+export interface Cat {type: \\"cat\\"; breed: \\"labrador\\" | \\"carlin\\" | \\"beagle\\"}
+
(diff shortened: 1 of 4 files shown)


src/scripts/tests/__snapshots__/import-open-api.test.ts.snap
```diff
@@ -12,6 +12,12 @@ export type Pet = NewPet & {id: number};
 
 export interface NewPet {name: string; tag?: string}
 
+export type CatOrDog = Cat | Dog;
+
+export interface Cat {type: \\"cat\\"; breed: \\"labrador\\" | \\"carlin\\" | \\"beagle\\"}
+
+export interface Dog {type: \\"dog\\"; breed: \\"saimois\\" | \\"bengal\\" | \\"british shorthair\\"}
```
TejasQ: Why the double escape?

fabien0102: It's a snapshot ^^ I'm not responsible of jest formatting 😁

TejasQ: I don't really care about the formatting. I'm just wondering if/how it actually works:

image

fabien0102: image

It's not a typescript file! It's a SNAPSHOT 😅

Issue #181, opened 2019-11-19
React Native support (issue with React peer dependency) 


owinter86:
**Describe the bug**
This currently does not work with new React Native projects (non expo) due to RN targeting a specific React Version (currently 16.9.0), so due to the react peer dependency it will install react@^16.8.4 and therefore cause the error shown below.

I can confirm there are no issues if we remove react as a peer dependency, which will ensure when using with React native that two versions of react will not be installed. But unsure if this is a satisfactory resolution.

**To Reproduce**
1. init new react native project (without expo)
2. install react restful
3. try and use any hook

**Expected behavior**
Should support the React Native Projects

**Screenshots**
<img width="540" alt="Screen Shot 2019-11-19 at 11 49 52" src="https://user-images.githubusercontent.com/2920419/69140411-c4359300-0ac2-11ea-9936-187ec838874a.png">

South-Paw:
@TejasQ Looks like this issue can be closed as well as PR has been merged.

fabien0102:
Seams to be fixed, please tell me if not 😉

PR #182, opened 2019-11-19
Update React peer dependency

owinter86:
# Why
This addresses issue here https://github.com/contiamo/restful-react/issues/181, this will ensure that future react native versions are supported and no longer installs multiple React versions, but unsure if this will have other implications

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,5 +104,8 @@
     "swagger2openapi": "^3.2.14",
     "url": "^0.11.0",
     "yamljs": "^0.3.0"
+  },	
+  "peerDependencies": {	
+    "react": "^16.8.5"
   }
-}
+} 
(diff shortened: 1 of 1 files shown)


TejasQ:
Hey @owinter86! Thanks so much for your pull request. I'm a little bit apprehensive about removing React as a peer dependency here because React is not listed under `dependencies` in `package.json` at all (this is a separate issue – it _should_ be because we're using it 😅).

What if instead of removing the peer dependency, we bump it to the lowest compatible version of React for web and native? I believe this is 16.9.x? In which case, `>=16.9.0` should be what we need.

That might be safer. What do you think?

owinter86:
Yep sounds good to me, do I need to update the PR

TejasQ:
Yes, because the PR removes it currently. 😬

owinter86:
I have updated to include peer dependency of react 16.8+ as I believe `>=16.9.0` will likely no work for the next RN update. I think you need React version that is compatible with hooks for this package to run, so hopefully this peer dependency change is ok.

TejasQ:
Great work!

PR #187, opened 2019-11-25
Improve the generated type of `additionalProperties`

fabien0102:
# Why

With a closer look to the specs, this part was a bit missaligned.

image

https://swagger.io/docs/specification/data-models/data-types/

We can now combine `item.properties` and `item.additionalProperties` as specify in the open-api. We also deal with `item.additionalProperties: true`


src/scripts/tests/import-open-api.test.ts
```diff
@@ -60,7 +60,7 @@ describe("scripts/import-open-api", () => {
       { item: { type: "array", items: { type: "integer" } }, expected: "number[]" },
       { item: { type: "array", items: { type: "customType" } }, expected: "any[]" },
       { item: { type: "object", properties: { value: { type: "integer" } } }, expected: "{value?: number}" },
-      { item: { type: "object" }, expected: "{}" },
+      { item: { type: "object" }, expected: "{[key: string]: any}" },
```
fabien0102: I also improve a bit this type, the `additionalProperties` is implicit in this part if I understand correctly the specs, so it's not an empty object.

src/scripts/import-open-api.ts
```diff
@@ -126,24 +127,29 @@ export const getObject = (item: SchemaObject): string => {
     return item.oneOf.map(resolveValue).join(" | ");
   }
 
+  // Consolidation of item.propeties & item.additionalProperties
```
TejasQ:
```suggestion
  // Consolidation of item.properties & item.additionalProperties
```

Issue #186, opened 2019-11-23
Lazy fetching of data will cause infinite loop / runaway effect


owinter86:
**Describe the bug**
With a lazy useGet, placing the refetch within an effect will cause an infinite loop of fetches due to the refetch function getting regenerated on each render.

**To Reproduce**
```
  const { refetch } = useGet(`/some-endpoint`, { lazy: true });

  React.useEffect(() => {
    refetch();
  }, [refetch]);
```

**Expected behavior**
I believe that refetch should be memoised and should not fail the dependency check in an effect. SO that you can safely refetch data in an effect with the refetch function as a dependency.

TejasQ:
Hi @owinter86! Thanks for your contribution! I have a question – why would you add `refetch` as a dependency to `useEffect` if you're don't want to rerender when it changes? Happy to memoize it on the library side, but this seems like an interesting pattern...

owinter86:
Hey @TejasQ , I totally agree that the above effect should not have refetch as a dependency, I simplified this example to show the issue I am having.

This is more of a developer experience improvement that an actual bug, as the hooks eslint rule will flag this if refetch is not included in the dependency array, this will be a bit problematic especially if you have an effect with many dependencies that you want to trigger a refetch. So you would have to disable this rule and remove refetch from the auto generated array. 

I also see that the the internal apis for setState, and a useReducer dispatch allow you to safely add these as dependencies without the problem of being recreated on re-render. Also some external libraries manage this as well, react-redux is useDispatch is a good example.

I just feel like that many users who rely on this eslint rule to either auto fill the dependency array or use it to flag missing dependencies will need to disable it any time they would like to have refetch in the effect.

I think that it will likely improve DX if the library can memoise this function so it can safely be used with the hooks eslint rule, and given there is no return value, and its just updating another reference, I dont think it needs to be recreated each update.

What do you think?

TejasQ:
While I agree that it will make people's lives easier and improve the DX to have these functions included in the `useEffect` dependency "watch list", I think it's quite misleading to just populate it with things inside the effect, even if we don't _explicitly_ want to watch them. It's kind of abusing the injection mechanism of `useEffect` and, in my opinion, an anti-pattern.

But we probably don't want people to have to think of that and allow them to focus on their tasks at hand instead of worrying about library topics.

I wonder what the rest of the team things? @stereobooster @fabien0102 any ideas?

owinter86:
[...]

PR #188, opened 2019-11-25
[WIP] Memoize refetch in useGet hook

owinter86:
# Why

Addresses https://github.com/contiamo/restful-react/issues/186, refetch is memoized against the same dependency array as the auto `fetchData()` effect within the package.

### DIFF
--- a/src/useGet.tsx
+++ b/src/useGet.tsx
@@ -194,11 +194,11 @@ export function useGet<TData = any, TError = any, TQueryParams = { [key: string]
 
   const abortController = useRef(new AbortController());
 
-  // refetch function  with the same dependency array as the fetchData deep effect
+  // combine refetch options defined at the component level to the fetchData request
   const refetch = (options: RefetchOptions<TData, TQueryParams> = {}) =>
     fetchData({ ...props, ...options }, state, setState, context, abortController);
 
-  // Memoise refetch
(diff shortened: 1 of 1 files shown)


owinter86:
Sorry, I changed my train of thought a few times with this, I think refetch should not depend on anything and be referentially equal on first initialisation, but since the fetchData function is extracted out of the hook, I think I need to create two refetch function to ensure none of the data passed to  _fetchData is stale, hence why there is refetch and refetchMemo.

tonthanhhung:
@owinter86, Could you give some sort description why did you close this PR ?, I think it's quite helpful in many cases

owinter86:
From what I recall, I closed this as its not a simple solution as I first thought (basic memoization of refetch), retry is returning the fetchData function which is abstracted out of the component, and relies on props and state to be passed to it as arguments. 

So I felt as though the fetchData abstraction would need to be removed and rewritten, so it was to not rely on having those arguments passed to it, I am not sure that this would be in line with how this package was setup, and unfortunately did not have capacity to continue with the PR.

Thankfully the testing suite is nice and robust and caught out those issues identified. 

There is probably a straight forward solution that I am missing, but this PR wont solve that. Thats why I closed it.

Sorry I couldn't help out more.

Issue #178, opened 2019-11-15
Param names containing dots are not supported


seanlaff:
Example spec:
```json
{
   "openapi": "3.0.0",
   "info": {
      "version": "1.0.0",
      "title": "Swagger Petstore",
      "license": {
         "name": "MIT"
      }
   },
   "servers": [
      {
         "url": "http://petstore.swagger.io/v1"
      }
   ],
   "paths": {
      "/pets": {
         "get": {
            "summary": "List all pets",
            "operationId": "listPets",
            "tags": [
               "pets"
            ],
            "parameters": [
               {
                  "name": "dot.delim.name",
                  "in": "query",
                  "description": "How many items to return at one time (max 100)",
                  "required": false,
                  "schema": {
                     "type": "integer",
                     "format": "int32"
                  }
               }
            ],
            "responses": {
               "200": {
                  "description": "A paged array of pets",
                  "headers": {
                     "x-next": {
                        "description": "A link to the next page of responses",
                        "schema": {
                           "type": "string"
                        }
                     }
                  },
                  "content": {
                     "application/json": {
                        "schema": {
                           "$ref": "#/components/schemas/Pets"
                        }
                     }
                  }
               },
               "default": {
                  "description": "unexpected error",
                  "content": {
                     "application/json": {
                        "schema": {
                           "$ref": "#/components/schemas/Error"
                        }
                     }
                  }
               }
            }
         }
      }
   },
   "components": {
      "schemas": {
         "Pet": {
            "type": "object",
            "required": [
               "id",
               "name"
            ],
            "properties": {
               "id": {
                  "type": "integer",
                  "format": "int64"
               },
               "name": {
                  "type": "string"
               },
               "tag": {
                  "type": "string"
               }
            }
         },
         "Pets": {
            "type": "array",
            "items": {
               "$ref": "#/components/schemas/Pet"
            }
         },
         "Error": {
            "type": "object",
            "required": [
               "code",
               "message"
            ],
            "properties": {
               "code": {
                  "type": "integer",
                  "format": "int32"
               },
               "message": {
                  "type": "string"
               }
            }
         }
      }
   }
}
```
<img width="538" alt="Screen Shot 2019-11-14 at 8 19 16 PM" src="https://user-images.githubusercontent.com/3194832/68909418-2df91a00-071c-11ea-85bb-87642d94d020.png">

TejasQ:
[...]

PR #190, opened 2019-11-28
fix: add double quotes for special properties

aisensiy:
# Why

For this issue https://github.com/contiamo/restful-react/issues/178 and using double quotes can solve the problem ([more info here](https://stackoverflow.com/questions/13661144/typescripts-hashtables-and-keys-with-a-dash)). I use a simple version regexp to test the `identifier` in javascript. More info can be found [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Working_with_Objects).


aisensiy:
Is there any one handle this?

fabien0102:
Published in `restful-react@9.4.2` 😉

PR #195, opened 2019-12-04
feat: set RestfulProviderContext as displayName for Context component

alfdocimo:
# Why

Add capabilities to display the name of the Provider and the Consumer as per #107


fabien0102:
@alfdocimo Published in `restful-react@9.4.1` 😉

Issue #180, opened 2019-11-18
async/promises for requestOptions in RestfulProvider


South-Paw:
### Is your feature request related to a problem?

We use aws-amplify library for auth. To get the current users JWT, we call `Auth.currentSession()` which is a `Promise<User>` where `User` contains a valid JWT (as the lib handles token refresh as well).

The problem is that `RestfulProvider`'s `requestOptions` does not allow for promises or async functions that return `Partial<RequestInit>`.

### Describe the solution you'd like

One of

* `requestOptions` supports async functions

* `requestOptions` supports `Promise<Partial<RequestInit>>` type

* another alternative workaround for this issue

### Describe alternatives you've considered

Current workaround I have in place is to store the JWT in Redux and connect the RestfulProvider to the store to get the JWT for requests.

However, this was an interim workaround and the JWT in Redux is not refreshed when the lib internally refreshes the JWT token - so we get 401's after the first token refresh.

### Additional context

* [Sample of code in solution one](https://user-images.githubusercontent.com/2154764/69096782-f7275900-0ab9-11ea-8cb2-e7add5fe44ae.png)
* [Sample of code in solution two](https://user-images.githubusercontent.com/2154764/69096769-ef67b480-0ab9-11ea-900b-724674b25eec.png)

Any help or feedback is appreciated as well - big thanks for improving the way I write React components 🙌

South-Paw:
@fabien0102 Hi there, thanks for the friendly reply!

If its reasonably straight forward as you say I'd be more than happy to have a go at implementing it.

Do you have any suggestions on where to start with this? I haven't yet had a chance to dig through the code, but maybe an indication of what files I'll need to touch to start the ball rolling

fabien0102:
@South-Paw For this kind of feature, a good start is probably to update the type here -> https://github.com/contiamo/restful-react/blob/master/src/Context.tsx#L21 (so you can see where typescript is complaining ^^)

And after add some unit tests over there -> https://github.com/contiamo/restful-react/blob/master/src/Get.test.tsx#L526 with a promise as `requestOptions` and see what you need to update to make the test happy 😁 

The logic is quite isolate, we have some `getRequestOption` functions (https://github.com/contiamo/restful-react/blob/master/src/Get.tsx#L207), logically with typescript and unit tests, you "just" have to follow the red. Don't hesitate to open a draft in github if you need any help 😉

South-Paw:
[...]

PR #199, opened 2019-12-12
Promise request options

South-Paw:
# Why

Second go at allowing for promised request options to be provided to components and hooks.

See #180 for more detail

### DIFF
--- a/src/Poll.tsx
+++ b/src/Poll.tsx
@@ -219,7 +219,7 @@ class ContextlessPoll<TData, TError, TQueryParams> extends React.Component<
     // If we should keep going,
     const { base, path, interval, wait } = this.props;
     const { lastPollIndex } = this.state;
-    const requestOptions = this.getRequestOptions();
+    const requestOptions = await this.getRequestOptions();
 
     let url = composeUrl(base!, "", path);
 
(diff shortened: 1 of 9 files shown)


src/Poll.tsx
```diff
@@ -343,13 +343,13 @@ function Poll<TData = any, TError = any, TQueryParams = { [key: string]: any }>(
   // Compose Contexts to allow for URL nesting
   return (
     <RestfulReactConsumer>
-      {contextProps => {
+      {async contextProps => {
         const contextRequestOptions =
           typeof contextProps.requestOptions === "function"
-            ? contextProps.requestOptions()
+            ? await contextProps.requestOptions()
```
South-Paw: I saw you didn't add an await to this one in your commits @fabien0102 - was that accidental or did you intend to not await this one?

src/Poll.tsx
```diff
@@ -343,13 +343,13 @@ function Poll<TData = any, TError = any, TQueryParams = { [key: string]: any }>(
   // Compose Contexts to allow for URL nesting
   return (
     <RestfulReactConsumer>
-      {contextProps => {
+      {async contextProps => {
```
South-Paw: The poll component needs some kind of rework to support this I think.

Placing this async here is what creates all the invariant errors we see in the tests - React tries to render the Promise object of this function

I'm unsure how you want to handle this?

South-Paw:
@fabien0102 look! it's all green 😄

second times a charm it seems!

fabien0102:
Published in `restful-react@9.5.0`

PR #212, opened 2020-01-12
Update code to export umd, cjs, esm etc

weyert:
# Why

In the past there has been tickets raised that the library isn't straight away usable with `create-react-app` and other projects. I have created a test app at: https://github.com/weyert/restful-react-demo

This PR contains a bunch of changes at first the contained in PR #211 but then also:

 * Added new `rollup`-config specifically for the tools scripts named `tools.rollup.config.js`
* Changed the locations of the build artefacts:
      - Tools scripts end up in `lib` 
      - Actually components end up in `dist`
 * Expose size snapshots of the generated files
 * Moved the test files to their dedicated folder named `__tests__`
 * Added the necessary `rollup`-plugins


seanlaff:
Psyched for this!

fabien0102:
@weyert Sorry, I need to rebase this one, you have some commits from https://github.com/contiamo/restful-react/pull/211 in conflict

Something like this should do the job to avoid conflict resolution 😉:
`git rebase --onto origin/master features/remove-operational-scripts features/just-use-rollup feature/`

src/util/__tests__/composeUrl.test.ts
```diff
@@ -1,4 +1,4 @@
-import { composePath, composeUrl } from "./composeUrl";
```
fabien0102: Why did you move everything inside a `__tests__` folder? 🤔

weyert: No, really reason, mainly it was late and I was having a feeling it was leaking into the final builds. I am happy to revisit and move them back if you want.

fabien0102: I usually prefer my tests aside the component (otherwise people forget about these 😬), but I get the leaking part, I have the same locally 😕 (I was playing a bit with tsdx today, to give a try and see if need all this extra configuration)

PR #215, opened 2020-01-13
Migrate to Tsdx

fabien0102:
# Why

Our project is not compiled to umd, and I saw this package mentionned somewhere in our issues ^^ (and as a curious person, I had to give a try 😁)

@weyert I would like to have your feedback on this one, since you tried the "manual" version in #212 

# Notes
- I had to fix all the tests types since I increase the typesafety.
- I still have a custom rollup for the scripts due to https://github.com/jaredpalmer/tsdx/issues/338

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -8,8 +8,5 @@ module.exports = {
     "react": {
       "version": "detect"
     }
-  },
-  "rules": {
-    "array-callback-return": "off"
   }
 }
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,15 @@
+module.exports = {
+  "extends": [
+    "react-app",
+    "prettier/@typescript-eslint",
+    "plugin:prettier/recommended"
+  ],
+  "settings": {
+    "react": {
+      "version": "detect"
+    }
+  },
+  "rules": {
+    "array-callback-return": "off"
+  }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -102,7 +102,3 @@ package-lock.json
 
 # examples artefacts
 examples/*.tsx
-
-# Files managed by operational-scripts
-tslint.json
-.prettierrc
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,5 @@
+{
+  "semi": true,
+  "printWidth": 120,
+  "trailingComma": "all"
+}
--- a/package.json
+++ b/package.json
@@ -102,6 +102,7 @@
     "ibm-openapi-validator": "^0.16.0",
     "inquirer": "^7.0.3",
     "lodash": "^4.17.15",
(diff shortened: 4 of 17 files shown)


weyert:
Hah, looks easier than my version. Let's go for this. I will see if I can publish the package with Verdaccio locally and then test if it works as expected in CRA

fabien0102:
@weyert You can try with `restful-react@9.7.1-3-gc0047cb` (publish as `canary` version) 😉

.eslintrc.js
```diff
@@ -0,0 +1,15 @@
...
+  ],
+  "settings": {
+    "react": {
+      "version": "detect"
+    }
+  },
+  "rules": {
+    "array-callback-return": "off"
+  }
+}
```
TejasQ:
```suggestion
}

```

.eslintrc.js
```diff
@@ -0,0 +1,15 @@
...
+    "prettier/@typescript-eslint",
+    "plugin:prettier/recommended"
+  ],
+  "settings": {
+    "react": {
+      "version": "detect"
+    }
+  },
+  "rules": {
+    "array-callback-return": "off"
```
TejasQ: Why?

fabien0102: Because we use this syntax and it was complaining (and don't see any good reason for this rule 😁)

ref: https://eslint.org/docs/rules/array-callback-return

TejasQ: If you forget to `return` inside `Array.prototype.map`, or `Array.prototype.reduce`, you end up with an array of `undefined`. Could be dangerous, no? 🤔

fabien0102: I can maybe give a try for `["error", { allowImplicit: true }]`

fabien0102: > If you forget to `return` inside `Array.prototype.map`, or `Array.prototype.reduce`, you end up with an array of `undefined`. Could be dangerous, no? 🤔

If you finish with an array of `undefined` and you commit this without noticing it, YOU are dangerous (everything should be tested ^^)

fabien0102: But looking a bit more where it was failing, I can just transform my `.map` and `.forEach` and everything follow, so let's go back on the "recommended" setup ^^

fabien0102:
New version here to test in real condition -> `restful-react@9.7.1-6-g0969004`

fabien0102:
image

Still need some work…

weyert:
Looks to be working for me :)

fabien0102:
Tested againts a fresh new CRA locally, everything follow!

weyert:
Sweet, same here! ✋ 💃

seanlaff:
Thanks for taking the time to do this- makes it much easier to consume :)

South-Paw:
Great job @fabien0102

PR #216, opened 2020-01-13
#205 Add skip flag

### LINKED ISSUE #205: [Question] How to create a conditional useGet ?
I try to achieve something like dependent `useGet` (and save the data to state also) after received the response from the first `useGet`
something similar to this https://github.com/zeit/swr#dependent-fetching
Is it possible now, and how would I do it ?
PS: 
I tried to take advantage of lazy & refetch, eg:
```
{data: user} = useGet({path: pathToGetUser})
{refetch: refecthProject} = useGet({
  path: pathToGetProjectWithUseId, 
  lazy:true, 
  resolve:data=>{
     //save data to state
  }})
if (user && user.id){
  refetchProject() 
}
``` 
That was my idea
but I stuck at how to compose the `pathToGetProjectWithUserId` is ugly and error prone like this
`/api/project/user/{user ? user.id ? -1}`

How would I do the above requirement nicer ?

seanlaff:
Fixes #205 

I added skip to both get and poll. Not 100% sure we want it on poll. Thoughts?

As a side note, I can't `yarn install` at the moment from the private npm repo- not sure if its just a transient thing or upstream config changed.
```
error An unexpected error occurred: "http://verdaccio.bowerytree.co:4873/handlebars/-/handlebars-4.7.0.tgz: connect ECONNREFUSED 51.15.61.114:4873".
```

### DIFF
--- a/src/useGet.test.tsx
+++ b/src/useGet.test.tsx
@@ -455,6 +455,28 @@ describe("useGet hook", () => {
     });
   });
 
+  describe("with skip", () => {
+    it("should not fetch on mount", async () => {
+      const children = jest.fn();
+      children.mockReturnValue(<div />);
+
+      const MyAwesomeComponent = () => {
+        const { data, error, loading } = useGet<{ oh: string }>({ path: "/", skip: true });
+
+        return children({ data, error, loading });
+      };
+
+      render(
+        <RestfulProvider base="https://my-awesome-api.fake">
+          <MyAwesomeComponent />
+        </RestfulProvider>,
+      );
+
+      await wait(() => expect(children).toBeCalledTimes(1));
+      expect(children).toHaveBeenCalledWith({ data: null, error: null, loading: false });
+    });
+  });
+
   describe("with base", () => {
     it("should override the base url", async () => {
       nock("https://my-awesome-api.fake")
--- a/README.md
+++ b/README.md
@@ -242,6 +242,20 @@ export default MyComponent;
 
 The above example will display your UI, and then load good boyes on demand.
 
+### Skip Fetching
+
+In cases where a parameter to a request may not yet be available, a request can be skipped with the `skip` boolean property. Skipping is supported by both `Get` and `Poll`.
+
+```jsx
+const { data: randomDogImage, loading } = useGet({
+  path: "https://api/user",
+  skip: id === undefined,
+  queryParams: {
+    userId: id,
+  },
+});
+```
+
 ### Response Resolution
 
 Sometimes, your backend responses arrive in a shape that you might want to adapt, validate, or restructure. Other times, maybe your data consistently arrives in a `{ data: {} }` shape, with `data` containing the stuff you want.
--- a/src/Get.tsx
+++ b/src/Get.tsx
@@ -102,6 +102,10 @@ export interface GetProps<TData, TError, TQueryParams> {
    */
   lazy?: boolean;
   /**
+   * Should we skip this request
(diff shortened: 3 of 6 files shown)


weyert:
Looks like I have some configuration wrong over here.

src/useGet.test.tsx
```diff
@@ -455,6 +455,28 @@ describe("useGet hook", () => {
...
+      };
+
+      render(
+        <RestfulProvider base="https://my-awesome-api.fake">
+          <MyAwesomeComponent />
+        </RestfulProvider>,
+      );
+
+      await wait(() => expect(children).toBeCalledTimes(1));
+      expect(children).toHaveBeenCalledWith({ data: null, error: null, loading: false });
```
fabien0102: I think your test doesn't test the correct thing here, nothing assert that fetch is not called here.

Can you try to revert the following change in `useGet.tsx`

```tsx
useDeepCompareEffect(() => {
     if (!props.lazy /** && !props.skip **/) {
      fetchData(props, state, setState, context, abortController);
    }
}
```

and see if your test is really catching the implementation 😉

seanlaff: This test follows the pattern of asserting the child rendered 1 time (as opposed to two, which would happen if there was a network request), which I copied from the `lazy` test. I agree  the naming of both test cases is slightly misleading, but I assume it was done this way because it was hard to assert nock *wasn't* called.

Do you know if there is a good pattern for that? Happy to add it to both tests if so

fabien0102: I will try to give a try locally today, I just want to be sure that we are testing the right thing (and this will not break without any red on CI 😁)

fabien0102:
Regarding a bit closer to the implementation, `skip` is just an alias for `lazy` 🤔 

I'm wondering if we really want to add this "duplication" or if we maybe want another behaviour for this `skip`

@TejasQ any feedback? (I need another point of vue ^^)

TejasQ:
I think you're right, @fabien0102. `skip` and `lazy` are the same thing, and since `lazy` is `boolean`, I think we're kind of reinventing the wheel in this case. 🤔 

I may be wrong but the [definition of lazy](https://github.com/contiamo/restful-react/blob/122b4d469d670fcc6b39155a78f9d120da41d192/src/Get.tsx#L101) is the same as [the expected behavior of `skip`](https://github.com/contiamo/restful-react/pull/216/files#diff-71e238669ebb35ceb38e93936fe04aedR666).

Maybe we can make `lazy` a little bit more explicit and meet everyone's requirements?

seanlaff:
@TejasQ  @fabien0102  Good suggestions. I had assumed `lazy` had a deeper tie in with the logic for `refetch()`, but under closer inspection I see that `skip` as I have it is just an alias for the same behavior.

One argument for 2 variables would be the case where you have a lazy get that you want to call `refetch()` for from many different places, however you want "bake in" the skip logic to the original `useGet()` call so you don't have to duplicate your gate to all of the places that call `refetch()`.

Is that a contrived example? Not entirely sure :) The being said, this MR would require a little more work to make that above statement true.

How would you like to proceed?

fabien0102:
Not totally sure about this one, for now I want to say that `lazy` do the job, I guess we can improve a bit the documentation instead of going deeper (another props also means more complexity and more potential bugs ^^)

Let's close this PR for now, since `skip` is just an alias for `lazy` 😁

seanlaff:
Sounds good 👍

PR #217, opened 2020-01-13
Update yarn.lock file to use public npm registry

weyert:
# Why

My test npm registry got part of the yarn.lock due global npm config on my side. Sorry about this!


fabien0102:
Indeed! Don't worry, it's also my fault, I should reviewed more carrefully ;)

PR #219, opened 2020-01-14
Export UseGetReturn and UseMutateReturn types from index

South-Paw:
# Why

Our app uses `UseGetReturn` and `UseMutateReturn` types but after `9.8.0` we were not longer able to import them without eslint errors. I've added them to the index exports which seems to have fixed it when locally tested in our app.

Also organised the `package.json` and added a few more checks to the `tsconfig.json`


src/scripts/ibm-openapi-validator.d.ts
```diff
@@ -15,7 +15,7 @@ declare module "ibm-openapi-validator" {
    * @param openApiDoc An object that represents an OpenAPI document.
    * @param defaultMode If set to true, the validator will ignore the .validaterc file and will use the configuration defaults.
    */
-  function validator(openApiDoc: any, defaultMode = false): Promise<ValidatorResults>;
+  function validator(openApiDoc: any, defaultMode?: any): Promise<ValidatorResults>;
```
fabien0102: Even if I need to remove this validator at some point (way to strict to be usable 😕), I'm curious about this change!

fabien0102:
Looks good, but I need to try locally ;)

src/scripts/ibm-openapi-validator.d.ts
fabien0102: I reverted the change, and nothing red so far, can you revert this?

South-Paw: I changed it back, weird that its no longer red 🤔

Only reason I fiddled with it was because it was red 😄

South-Paw:
@fabien0102 reverted that ibm validator type change 👍

package.json
```diff
@@ -45,7 +45,6 @@
     "build": "run-p build:*",
     "build:project": "tsdx build",
     "build:bin": "tsc && rollup -c rollup.config.js",
-    "prepare": "tsdx build",
```
South-Paw: @fabien0102 I think this caught out whoever did the publish of `9.8.0` so I've removed it from the scripts.

See: #222

fabien0102: Correct, but this should be `npm run build` instead, so we are sure to not forget to build before publishing (oh yes, this happens so easily! 😁)

I will this directly on master before publishing 😉

fabien0102:
@South-Paw Published on `restful-react@9.9.0` 😉

Issue #231, opened 2020-02-03
UTF-8 charset in response content causing the generator to replace all return types with void


ChristianIvicevic:
**Describe the bug**
When defining the responses and using the `application/json;charset=UTF-8` mimetype the code generator fails to correctly type all requests and replaces the return types with `void`.

**To Reproduce**
Create an OpenAPI 3 schema file with this endpoint:
```
paths:
  /api/activities:
    get:
      operationId: getActivities
      responses:
        200:
          description: OK
          content:
            application/json;charset=UTF-8:
              schema:
                type: array
                description: Unordered list of activity names
                items:
                  type: string
```

**Expected behavior**
I expected the `useGetActivities` hook to return a `string[]`, but it is typed as `void`. Replacing `application/json;charset=UTF-8` with `application/json` fixes this issue.

**Desktop (please complete the following information):**
 - OS: macOS Catalina 10.15.2

fabien0102:
Nice catch, should be quite easy to fix, do you want to take this issue?

fabien0102:
We need to improve a bit this part to be more permissive

https://github.com/contiamo/restful-react/blob/48447507921c60590b30553c740e49cdd64cbfb1/src/scripts/import-open-api.ts#L198-L208

fabien0102:
Was raining outside 🙄 So I fixed it ^^ Should be merge tomorrow 😉

PR #232, opened 2020-02-03
Deal with charset

fabien0102:
# Issue

#231

# Why

We didn't deal with more advanced content type, this should be way more flexible.

### DIFF
--- a/src/scripts/import-open-api.ts
+++ b/src/scripts/import-open-api.ts
@@ -195,12 +195,14 @@ export const getResReqTypes = (
       if (isReference(res)) {
         return getRef(res.$ref);
       } else {
-        if (res.content && res.content["application/json"]) {
-          const schema = res.content["application/json"].schema!;
-          return resolveValue(schema);
-        } else if (res.content && res.content["application/octet-stream"]) {
-          const schema = res.content["application/octet-stream"].schema!;
-          return resolveValue(schema);
+        if (res.content) {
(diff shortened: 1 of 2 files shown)


src/scripts/import-open-api.ts
```diff
@@ -195,12 +195,14 @@ export const getResReqTypes = (
...
+          for (let contentType of Object.keys(res.content)) {
+            if (contentType.startsWith("application/json") || contentType.startsWith("application/octet-stream")) {
+              const schema = res.content[contentType].schema!;
+              return resolveValue(schema);
+            }
+          }
+          return "void";
         } else {
           return "void";
         }
```
mpotomin: Since we are touching this part of the code, let's improve the readability a little bit by decreasing the amount of levels of code indentation. I suggest:

```typescript
export const getResReqTypes = (
  responsesOrRequests: Array<[string, ResponseObject | ReferenceObject | RequestBodyObject]>,
) =>
  uniq(
    responsesOrRequests.map(([_, res]) => {
      if (!res) {
        return "void";
      }

      if (isReference(res)) {
        return getRef(res.$ref);
      }

      if (res.content) {
        for (let contentType of Object.keys(res.content)) {
          if (contentType.startsWith("application/json") || contentType.startsWith("application/octet-stream")) {
            const schema = res.content[contentType].schema!;
            return resolveValue(schema);
          }
        }
        return "void";
      }

      return "void";
    }),
  ).join(" | ");
```

PR #237, opened 2020-02-07
fix: codegen for array of enums

vkbansal:
# Why

For the following object
```yml
Request:
  properties:
    actions:
      type: array
      items:
        type: string
        enum:
          - create
          - view
          - update
          - delete
```
produces

```ts
interface Request {
  actions: 'create' | 'view' | 'update' | 'delete''[]
}
```

but expected result is

```ts
interface Request {
  actions: ('create' | 'view' | 'update' | 'delete'')[]
}
```

### DIFF
--- a/src/scripts/import-open-api.ts
+++ b/src/scripts/import-open-api.ts
@@ -101,7 +101,7 @@ export const getRef = ($ref: ReferenceObject["$ref"]) => {
  */
 export const getArray = (item: SchemaObject): string => {
   if (item.items) {
-    if (!isReference(item.items) && (item.items.oneOf || item.items.allOf)) {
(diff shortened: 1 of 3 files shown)


fabien0102:
Nice catch! 👌 Can you add a unit test to be sure we don't broke this in the future?

vkbansal:
Can you point me where to add the test? Thanks.

vkbansal:
@fabien0102 I've added unit test for the scenario

vkbansal:
@fabien0102 Can you please give me a rough timeline, by when this will be published?

fabien0102:
@vkbansal Thanks for the reminder 😁 I'm currently in vacation so I totally forgot 🙄 

This is published in `v9.11.1`

vkbansal:
Thanks!

PR #239, opened 2020-02-11
Improve the publish flow

fabien0102:
# Why

Multiple reasons:
 - I need to have access to my computer to publish on npm
 - the `*.test.*` was includes in the npm package
 - some people don't want the cli part of the package (security audit, waste of time)

To answer to this, I propose the following solution/flow:
- Bump the version in the package.json
- On merge to master -> `v9.11.1` and `v9.11.1-without-cli` are built and published

Why this stange flow and not two separate packages?
 - Monorepo will change the entire repo architecture (and I'm lazy ^^)
 - I like to have only `restful-react` to install (with the cli that is for me one of the best feature of this package)
  - With this setup, everybody should be happy, since everything is automated

## Side note
The changelog will not follow… Still need to find a solution for this, I'm not really fan of our current changelog generator. I'm thinking to migrate to https://www.npmjs.com/package/@changesets/cli in the future, in the midtime, github and PR should be enough 🙄


TejasQ:
Why don't we have `-with-cli` be the option and ship without CLI by default?

fabien0102:
> Why don't we have `-with-cli` be the option and ship without CLI by default?

Because to me, the real power of restful-react is the CLI 😉 and this doen't break the API also.

PR #240, opened 2020-02-18
Support `queryParamStringifyOptions`

fabien0102:
# Why

Follow up of #163

This is to allow to pass down `queryParamStringifyOptions` to `qs`


jakewtaylor:
Glad to see this got picked up! Sorry for abandoning in the first place 😅

Issue #242, opened 2020-03-08
Gatsby static HTML build fails due to dependency on browser API AbortController


mbrookson:
**The problem**
I've recently ported my personal site into a Gatsby application to try it out. I love restful-react, but I've hit an issue where the Gatsby static HTML build fails due to restful-react hooks depending on AbortController at the point when the hook is run, rather than when the HTTP request is made. Gatsby suggest any interactions with browser APIs be run inside a `useEffect` so that they run when the browser binds to React, rather than at build time. However, this is happing in a hook so I cannot move it to inside a `useEffect` as it breaks the rules of hooks!

**Potential solution**
Haven't looked at the restful-react source code yet but it seems like creating an instance of AbortController is being done prematurely and should be done at the time of calling the `mutate` function exposed by `useMutate`, for example, rather than when `useMutate` is first run.

**Alternative solutions**
- Allow the restful-react hooks to be fired in a 'mock' way, which stubs them when not running in the browser. Not sure I think this is a good solution though, but would avoid any server-side issues.
- Only allow AbortController and other browser APIs to run in the browser, but still allow the fetching of data without that added functionality when running in a non-browser environment?

**Example usage**
![image](https://user-images.githubusercontent.com/20131783/76164348-1a981180-6146-11ea-8513-9ec0bf3f2e39.png)

**Link to Gatsby docs**
[https://www.gatsbyjs.org/docs/debugging-html-builds/](https://www.gatsbyjs.org/docs/debugging-html-builds/)

**Output of Gatsby build error**

![image](https://user-images.githubusercontent.com/20131783/76164292-a2315080-6145-11ea-9148-7d58b5761afe.png)

fabien0102:
Hi, thanks for the feedback!
This should not be too hard to fix, just make sure than AbortController exists before using it should do the trick.
I'm currently stuck at the hospital (16m free fall at climbing 😅) so I can't really take this issue now, but if you want to open a PR, I can review it 😉

mbrookson:
Hi. I'm working on a fix but want to test locally with my Gatsby build to make sure it's working, however I'm getting an error when running `yarn build` or `yarn watch`. Any ideas?

![image](https://user-images.githubusercontent.com/20131783/76168178-14655d80-6165-11ea-93e8-2ef8fee1f142.png)

fabien0102:
Never had this locally, very strange and no ideas, sorry...

mbrookson:
[...]

PR #243, opened 2020-03-08
Extract usage of AbortController into own hook to check if it is available

mbrookson:
This provides a hook that works regardless of whether AbortController API is available, for example if the code is running in the browser vs in static html build process.

Fix for https://github.com/contiamo/restful-react/issues/242

Would appreciate another person's manual verification if possible!


src/useAbort.ts
```diff
@@ -0,0 +1,25 @@
...
+export function useAbort() {
+  let abortController;
+
+  try {
+    abortController = new AbortController();
+  } catch {
+    abortController = null;
+  }
+
+  const [instance, setInstance] = useState(abortController);
```
fabien0102: I would prefer to use useReference here, to avoid useless re-render 😉

mbrookson: I did think that, but that means this hook has to expose the whole AbortController `instance`. If I'm not mistaken, if it's a reference and it gets a new `current` value set, the exposed `signal` won't get updated to the new one wherever it's consumed. The reason I wanted to exposed a wrapper like this is so the consuming code doesn't have to have checks everywhere to see if the instance is undefined.

Not sure if there's another/better solution? Or how I could in fact use `useRef` but still solve the problem in a similar way?

mbrookson: @fabien0102 @TejasQ Could you shed any more light on this? 🙂 I'm not convinced the usage of `useState` is a big problem here since it would only cause a re-render if `abort` is called, and a re-render would also currently happen when the request resolves anyway. Would be interested to know if there is another solution though, and keen to be able to use this library with my Gatsby site!

fabien0102: Logically you can expose what ever you want, nothing force you to expose the entire ref 😉
Not really sure of re-render here, I need a computer to try ^^

Can you tell me if the tests are passing, for a strange reason Travis seams blocked...

If yes, you can do/push a "npm version patch" so you will have a new npm package when we merge this PR 😉

stereobooster: Not sure about `useState` usage, but you can improve code a bit like this

```ts
const [instance, setInstance] = useState(() => {
  try {
    return new AbortController();
  } catch {
    return null;
  }
});
```

mbrookson: Apologies, it does look like the test for `useGet` are failing. I'll take a look and make the suggested change above as it's a lot cleaner 👍

fabien0102: Don't worry, this is why we have some unit tests 😁

mbrookson:
@fabien0102 Looks like Travis CI might need a manual restart? Tests should pass again now but seems to be stuck still.

TejasQ:
DO NOT MERGE THIS because I need to show this to @github

TejasQ:
FYI –– I disabled Travis and run tests on Netlify now so if Netlify is green on future PRs, it means tests pass.

mbrookson:
> DO NOT MERGE THIS because I need to show this to @github

Is this because of a Travis issue?

TejasQ:
> Is this because of a Travis issue?

@mbrookson, no. It's a GitHub issue. They don't consider PRs approved on mobile devices even though they are.

Also, while the tests pass, it seems we're leaking memory in `Get` according to this PR:

image

TejasQ:
I don't see that memory leak locally though and it may not be a big deal. I'm more concerned with GitHub. 

The issue I'm trying to help solve is this one. Notice how even though the PR is approved, it will not permit a merge.

ezgif com-resize (2)

Unfortunately, this GIF and video reproduction isn't enough for GitHub to realize it's an issue. They asked for a _live PR_, and this PR happens to serve that purpose. It'll lose that purpose when we merge it, hence my "do not merge".

TejasQ:
Ah, nevermind. All passing checks render it mergeable on my phone. If we had one pending check, the issue would have persisted.

Thanks for your contribution, @mbrookson!

CHANGELOG.md
```diff
@@ -4,460 +4,460 @@ All notable changes to this project will be documented in this file. Dates are d
 
 Generated by [`auto-changelog`](https://github.com/CookPete/auto-changelog).
 
-#### [v9.12.0](https://github.com/contiamo/restful-react/compare/v9.11.1...v9.12.0)
+#### [v9.12.1](https://github.com/mbrookson/restful-react/compare/v9.11.1...v9.12.1)
```
stereobooster: we missed this one

mbrookson: Oops! That's a side-effect of running `npm version patch` on my local machine. 😱

fabien0102: Not a big issue ^^

Issue #173, opened 2019-11-14
Provide access to response headers


ysgk:
Hi, thanks for this library :)

**Is your feature request related to a problem? Please describe.**
I need access the `X-Total-Count` header for pagination, but no means seems to be provided.

**Describe the solution you'd like**
I want to access response headers. I think [processResponse](https://github.com/contiamo/restful-react/blob/da8882f9582fe5b98db6870a481924abc68b6275/src/util/processResponse.ts) should return `headers` too.

fabien0102:
@ysgk Do you want to give a try for this feature?

ysgk:
@fabien0102 Yes, I will make a PR.

ysgk:
I noticed that `GetState` already has `response` as its property, but it seems to be not implemented yet. Do you have any plan to implement that? If so, `headers` can be easily accessed from it.

https://github.com/contiamo/restful-react/blob/b1d4c8295225721d9ffc6f80eee54f47c4a0a5d8/src/Get.tsx#L133-L138

TejasQ:
@ysgk, that'd be an amazing contribution from you 😻

ysgk:
I think `data` or `error` and their header should be coupled tightly, otherwise we might get the wrong header info as `data` or `error`. That would be tough 😅

TejasQ:
Not sure I agree: we can clear response `headers` on every request initialization and reset them on every request completion – then they're always up to date and in sync.

fabien0102:
Closed by #254 

Please tell me if this resolution works for you, if not, let's find a better API all together!

Issue #172, opened 2019-11-08
RestfulProvider resolve property issue


yeldisbayev:
**Describe the bug**
Hi, thanks first of all for your work. I have trouble on resolving server response on RestfulProvider. Its doesn't work for me.

[Code example](https://github.com/yeldisbayev/restful-react-example/blob/master/src/index.tsx)

Also I have a trouble with mocking server(I use Swagger Hub Virtual server), especially with identification codes. As I understand, RestfulProvider resolve prop function is global for any requests. It will be convenient if we have opportunity to write separate resolve functions by GET, POST, PUT, DELETE request types. This changes may resolve mocking problem of response body id filed mocking on the project. For example(todo list app): we have TodoList, TodoItem, CreateTodoForm components. 
- TodoForm
Here we create new todo, and get from response new todo item with const id and add new todo item to the TodoList. After every request we will get same todo with same id, it broke React's list mapping TodoItem key property uniqueness requirement. If we can write seperate resolve functions for request method types, and after every post request generate new unique id on every response.  

- TodoItem
Here we changing todo status(done or not). It will be good, if resolve function gets request body(changed todo with id), response body. Swagger generated server  also return static response body maybe with another id. If we can get request body, we may set necessary id and status. We can write one function on RestfulProvider resolve prop especially only for PUT requests.

[Implementatin example](https://github.com/yeldisbayev/restful-react-example/blob/master/src/examples/index.tsx)

TejasQ:
Hey @yeldisbayev! Thank you for opening this issue! Apologies for the late response!

I think this update to the global `resolve` function could actually be quite useful. I'd love to spend some time on this, but I'd first like to offer you the opportunity to implement it and open a pull request. Is that something you'd be interested in?

Alternatively, each component (`Get`/`Mutate`) and hook (`useGet`/`useMutate`) also supports a `resolve` function that might give you the functionality you're looking for, since `useMutate` is used for non `GET` HTTP verbs. Perhaps that might be worth exploring first?

Happy to hear your thoughts and ideas.

fabien0102:
Closed by #254 

Please tell me if this resolution works for you, if not, let's find a better API all together!

Issue #171, opened 2019-11-07
Support request/response interceptors


schneck:
Hi there,

thanks for this libary, looks really nice. I have a use case however I'm not sure how it could be accomplished:
In our project, we're using Keycloak as auth provider, and we are working with JWT auth/refresh tokens. Whenever a request fails with an 401, we request a fresh auth token, and while this request is running, we buffer all incoming requests until the new token arrived. After that, the buffered requests are performed.

I did not find any possibility to intercept the request/response cycle somehow. Is there any option I didn't see?

Thanks!

josvegit:
I had the same question, I realized you can do it in the RestfulProvider, with the **resolve**  and **requestOptions** callbacks.

i.e before you return resolved data => do something and before returning requestOptions, do something

Since RestfulProvider is global for all requests it works like interceptors. I have not tried it fully yet but it should work

josvegit:
Think there is an error callback also, where you should be able to handle http-error codes, not sure about that one though

smilexu:
I have very similar usage, in my react project, I need to get user data and then put into Context. The `resolve` solution works great for me, thanks @josvegit .

fabien0102:
Closed by #254 

Please tell me if this resolution works for you, if not, let's find a better API all together!

smilexu:
@fabien0102 looks good , the onResponse callback on useGet level work great for me. Did you publish the new version? I am using 11.0.0, but seems not have this props on useGet

fabien0102:
I'm on it :wink:

smilexu:
Excellent, thanks for your effort on this library 👍 great job

PR #254, opened 2020-03-27
Added onRequest & onResponse in RestfulProvider

ahansaary:
# Why

## Problem
I have a problem with handling some global actions with request and response to handle refresh token on authentication cycle, for example I need to check on every request if the access token is expired or not and replace the Authorization header value with refresh token instead of the access token and on the response of this request I need to get the new access token and save it globally.

## Solution
Added two new triggers **onRequest**, **onResponse** like **onError** trigger and added call for these triggers when initializing the request and when get a new response on all library components and hooks.

## Use Cases
- handle any staff needed to be handled globally depend on request or response objects.
- listen to all requests and responses globally for debugging and logging purposes.
- I guess this PR has a possible solution for these three issues #171, #172, #173

Thank you for your great library and your great effort :pray:

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,14 @@ export interface RestfulReactProviderProps<T = any> {
    * to deal with your retry locally instead of in the provider scope.
    */
   onError?: (err: any, retry: () => Promise<T | null>, response?: Response) => void;
+  /**
+   * Trigger on each request.
+   */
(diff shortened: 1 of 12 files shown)


fabien0102:
Ooops, totally missed this notification 🙄 will review this after lunch 😉

PR #257, opened 2020-04-03
Support FormData

### LINKED ISSUE #168: Bug with sending FormData as request body
**Describe the bug**
Sending a multipart/form-data request with FormData as a body is not possible currently.

When you use FormData as a body, the `Content-Type` header should be set to `multipart/form-data; boundary=<BOUNDARY_NAME>`. The browser usually handles this itself - not specifying a `Content-Type` header causes the browser to apply one itself.

The problem is that `useMutate` is always guessing a `Content-Type` header, based on the request body - meaning the browser won't set the header, and thus the request is causing errors on our backend (`System.IO.InvalidDataException: Missing content-type boundary.`). I'm fairly certain this isn't unique to us, but it is a weird edge case.

As far as I can tell, there's no way you can specify the boundary name manually, so the only option would be to not set the `Content-Type` header when the request body is an instance of `FormData`.

**To Reproduce**
```tsx
const useFileUpload = () => {
  const { mutate, loading } = useMutate({
    path: '/file/upload/path',
    verb: 'POST',
  });

  return {
    loading,
    upload(file: File) {
      const formData = new FormData();
      formData.append('file', file);

      return mutate('', { body: formData });
    },
  };
}
```

On our C# API (I'm not entirely sure what our backend stack is), this request fails - because of the `Content-Type` header set in `useMutate`.

**Expected behavior**
If request body is an instance of FormData, no 'Content-Type' header should be set by restful-react, it should be left undefined so that the browser can handle it.

**Screenshots**
Example request payload:
image

The 'Content-Type' header, by default, is set to 'text/plain', overriding it manually with 'multipart/form-data' causes our API error because the boundary name has not been configured. It should be set to `multipart/form-data; boundary=<BOUNDARY_NAME>`

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome / Opera
 - Version: latest

Atrox:
Allows you to pass FormData objects to useMutate. 

Example:
```javascript
const formData = new FormData()
formData.append('image', file)

const response = await mutate(formData)
```

fixes #168

### DIFF
--- a/src/useMutate.tsx
+++ b/src/useMutate.tsx
@@ -84,13 +84,21 @@ export function useMutate<
 
       const options: RequestInit = {
         method: verb,
-        headers: {
-          "content-type": typeof body === "object" ? "application/json" : "text/plain",
-        },
       };
 
+      // don't set content-type when body is of type FormData
+      if (!(body instanceof FormData)) {
+        options.headers = { "content-type": typeof body === "object" ? "application/json" : "text/plain" };
+      }
+
       if (!isDelete) {
-        options.body = typeof body === "object" ? JSON.stringify(body) : ((body as unknown) as string);
+        if (body instanceof FormData) {
+          options.body = body;
+        } else if (typeof body === "object") {
+          options.body = JSON.stringify(body);
+        } else {
+          options.body = (body as unknown) as string;
+        }
       }
 
       const signal = getAbortSignal();


fabien0102:
Published on `11.2.0` :wink: thank again for your contribution!

andriy-ilin:
@Atrox Maybe you can do the same for Mutation component?

Issue #261, opened 2020-04-20
processResponse support blob & Co


cbek:
**Is your feature request related to a problem? Please describe.**
Currently, processResponse supports only JSON or text as response object.
The fallback is a text object. 
In few cases you would like to get a Blob or arrayBuffer.

**Describe the solution you'd like**
If the content-type is "application/json" or "text/plain" behavior will not change.
All other cases will return whole response object.
Check the PR:
https://github.com/contiamo/restful-react/pull/262

**Describe alternatives you've considered**
-

**Additional context**
-

PR #262, opened 2020-04-20
#261 - processResponse support blob & Co

cbek:
- if the content-type is "application/json" or "text/plain" behavior will not change.
- all other cases will return whole response object.

# Why
See: https://github.com/contiamo/restful-react/issues/261

### DIFF
--- a/src/util/processResponse.ts
+++ b/src/util/processResponse.ts
@@ -14,9 +14,21 @@ export const processResponse = async (response: Response) => {
         responseError: true,
       };
     }
+  } else if ((response.headers.get("content-type") || "").includes("text/plain")) {
+    try {
+      return {
+        data: await response.text(),
+        responseError: false,
+      };
+    } catch (e) {
+      return {
+        data: e.message,
+        responseError: true,
+      };
+    }
   } else {
     return {
-      data: await response.text(),
+      data: response,
       responseError: false,
     };
   }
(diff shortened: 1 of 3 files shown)


src/util/processResponse.ts
```diff
@@ -14,9 +14,20 @@ export const processResponse = async (response: Response) => {
...
     }
+  } else if ((response.headers.get("content-type") || "").includes("text/plain")) {
+    try {
+      return {
+        data: await response.text(),
+        responseError: false,
+      };
+    } catch (e) {
+      return {
+        data: e.message,
```
fabien0102, line 25: You forgot the `responseError: true` 😉

cbek:
@fabien0102 
Better?

fabien0102:
No ideas why the tests are not running, I will try this WE, don't have any computer where I am... (And I will publish a new package as soon as merge this PR 😉)

Thanks again for your contribution 😃

cbek:
> No ideas why the tests are not running, I will try this WE, don't have any computer where I am... (And I will publish a new package as soon as merge this PR 😉)
> 
> Thanks again for your contribution 😃

Cool.
Thanks.

fabien0102:
image

Two tests are failing, we need to check if they are still relevant or not 😉

cbek:
All tests should pass now.

PR #271, opened 2020-06-07
Openapi null type

cmaus-cs:
# Why

I encountered a problem where the openapi spec had a nullable type defined as

```
oneOf:
  - $ref: 'someRef'
  - type: null

```

This was translated into ```SomeRef | any```  because the type "null" was not supported explicitly as a scalar type

### DIFF
--- a/src/scripts/import-open-api.ts
+++ b/src/scripts/import-open-api.ts
@@ -60,6 +60,9 @@ export const getScalar = (item: SchemaObject) => {
     case "array":
       return getArray(item) + nullable;
 
+    case "null":
+      return "null";
+
     case "string":
     case "byte":
     case "binary":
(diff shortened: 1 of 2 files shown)


fabien0102:
Hello @cmaus-cs 👋 

Thanks for you contribution, it's always appreciated! but… `type: null` is actually not valid in open-api 🙄 

A screenshot/link of the specification for this part:
image
https://swagger.io/specification/#data-types

In short, you need to add `nullable: true` on your type, so something like this with your example
```yml
components:
  schemas:
    someRef:
       type: object
       nullable: true
       properties:
          ...
```

cmaus-cs:
Thanks for the suggestion, but I cannot rewrite the spec.

Please note that in v3.1 of the openapi-spec nullable is dropped completely and its usage is supposed to be replaced by using the the type "null". 

https://github.com/OAI/OpenAPI-Specification/blob/v3.1.0-dev/versions/3.1.0.md

fabien0102:
Indeed! I missed this update of open-api, with this in mind, your PR make totally sense 😃

fabien0102:
`v14.0.0` incomming for you 😉

fabien0102:
image

or not… sorry, I need to debug this…

PR #275, opened 2020-06-22
Refetch when provider props change

### LINKED ISSUE #272: Changes to requestOptions do not trigger refetch
We use `requestOptions` to set headers for authorization and tenancy info, but refetches do not get triggered when `requestOptions` changes, leading to stale state.

There is no check for comparing the previous vs new `requestOptions` like there is for `resolve`.
https://github.com/contiamo/restful-react/blob/61604ee50626b3c910224225cf4dd92cf4b260b0/src/Get.tsx#L191-L205

seanlaff:
`Get` and `UseGet` refetch when `requestOptions`, `base`, `parent`, or `queryParams` change on the provider.

Closes #272


fabien0102:
image

I can't merge on github… they seams to have a bug with the merge buttonn…, I will merge this old school style ^^

PR #276, opened 2020-06-24
Fix path params

fabien0102:
# Why

`pathParams` was forgot in the watching list, so the hook was not automatically fetching on props change


src/useGet.tsx
```diff
@@ -249,7 +249,17 @@ export function useGet<TData = any, TError = any, TQueryParams = { [key: string]
...
     };
-  }, [props.lazy, props.mock, props.path, props.base, props.resolve, props.queryParams, props.requestOptions, abort]);
+  }, [
+    props.lazy,
+    props.mock,
+    props.path,
+    props.base,
+    props.resolve,
+    props.queryParams,
+    props.pathParams,
```
fabien0102, line 259: This is the one added

PR #282, opened 2020-07-08
Fix build (babel helper-compilation-targets workaround)

amacleay-cohere:
This is a bug in babel: https://github.com/babel/babel/issues/11216
It is closed but points to an issue in node: https://github.com/nodejs/node/issues/32107
The issue in node is closed as not an issue, so it seems as though babel needs to fix.

Also, check that build works in PRs.

# Why
The master build is failing and this package is not publishing


amacleay-cohere:
I don't actually know for sure this will work! I was hoping that I would see the github actions run present in the PR to validate.

fabien0102:
Let's try! 😁 (and thanks a lot, I really did have the motivation to dig into this issue ^^)

fabien0102:
WORKS 🎉 Thanks a lot again 💯

PR #284, opened 2020-07-15
Rework mutation error and data handling

### LINKED ISSUE #235: Throwing HTTP failure codes has unfortunate consequences
Thanks for making `restful-react`! It's a great library and tool (I'm using it to generate code from an OpenAPI spec). 

I really like how `useGet` works: 
1. The promise never rejects, so you can `await` on it safely without it throwing an error
2. It's easy to check if `data` or `error` is defined, and then act on it. 
(If I was to change anything, there are some cool things you can do with conditional types, so that the compiler can automatically infer that `data` **or** `error` is defined, but **never* both)  

However, `useMutate` has some drawbacks:
1. The promise can throw, meaning that `await` is not safe without appending an empty `.then` clause: 
    ```typescript
    await useMutate("path", props).catch(() => { /* pass */})
    ```
    Furthermore, even though the `error` field from `useMutate` is of type `GetDataError<TError>`, the compiler cannot infer that in the promise rejection clause, because failed promises can contain anything.
2. There's no `data` field, to accompany `error`. That means that I'll have to use two mechanisms: `await` on `mutate` to read the result of my operation, and check `error` for any failures. 

I'd think it would make a lot of sense for `useMutate` to be more similar to `useGet`: 
1. The promise returned by `mutate` should never reject, and also resolve to `void`
2. A new field `data: TData | null` should be added, to mirror the existing `error`. 

Do you have any thoughts on this? I'd be happy to contribute some code to implement this functionality.

torkelrogstad:
Closes #235 

We make two changes to mutation:
    1. Don't throw on non-OK HTTP responses
    2. Don't return data, but set state

This is in order to mimic other data fetching options. For more
motivation on this, see this GitHub issue: https://github.com/contiamo/restful-react/issues/235

We also do some minor cleanup of types to avoid repetition and slighly
inconsistent types.

**Up for discussion:**
Shortly after finishing this code I got a new idea: what if errors that happen on **our** end are thrown, whereas errors that happen on the **server** end are paased in the `error` object returned from `useMutate`? This would mean: 

1. If the server returns with 4xx or 5xx we do the same thing as is happening in this code
2. If we fail to fetch (all the `catch`es after `fetch(...)`) we throw the error. From my understanding, the only times `fetch` throws is when it encounters serios networking errors, such as having no internet connection, running into CORS issues, or SSL issues. These events are truly exceptional, and since `fetch` throws them I think it's OK that we also throw them.
3. If the `resolve` function throws an error, we forward the error. 

This accomplishes a few things: 

1. We forward any errors caused by `fetch` or the users own code
2. OpenAPI (Swagger) works really nice with this approach: all `error`s from `useMutate`, `useGet`, etc get their proper typings from the OpenAPI spec files. Typing the errors from `resolve` or `fetch` on the other hand is much more difficult, since these events are not part of what OpenAPI describes. 

---

Let me know if it is hard to understand what I'm describing here, and I can see I can code up a prototype. This proposal would be a bit more work than what's here right now, and I'd like to get your guys' thoughts before spending a lot of time on it.

### DIFF
--- a/src/Context.tsx
+++ b/src/Context.tsx
@@ -1,8 +1,8 @@
 import noop from "lodash/noop";
 import * as React from "react";
-import { ResolveFunction } from "./Get";
+import { ResolveFunction, GetDataError } from "./types";
 
-export interface RestfulReactProviderProps<TData = any> {
+export interface RestfulReactProviderProps<TData = any, TError = TData> {
   /** The backend URL where the RESTful resources live. */
   base: string;
   /**
@@ -26,15 +26,7 @@ export interface RestfulReactProviderProps<TData = any> {
    * Depending of your case, it can be easier to add a `localErrorOnly` on your `Mutate` component
    * to deal with your retry locally instead of in the provider scope.
    */
-  onError?: (
-    err: {
-      message: string;
-      data: TData | string;
-      status?: number;
-    },
-    retry: () => Promise<TData | null>,
-    response?: Response,
-  ) => void;
+  onError?: (err: GetDataError<TError>, retry: () => Promise<TData | null>, response?: Response) => void;
   /**
    * Trigger on each request
    */
@@ -53,7 +45,7 @@ export interface RestfulReactProviderProps<TData = any> {
 export const Context = React.createContext<Required<RestfulReactProviderProps>>({
   base: "",
   parentPath: "",
-  resolve: (data: any) => data,
+  resolve: data => data,
   requestOptions: {},
   onError: noop,
   onRequest: noop,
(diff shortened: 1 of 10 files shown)


fabien0102:
wooo, nice job, I will need some time to try it locally to see the implication of this breaking change 😉 I'm a bit short in time right know but definitely looking forward for this PR 👍

torkelrogstad:
Have you had an time to look at this? I understand you're probably swamped in a ton of other things, but hope you'll be able to get to this eventually

torkelrogstad:
@fabien0102 any input on this?

fabien0102:
> @fabien0102 any input on this?

Totally forgot about this and totally under water to be honest 🙄 

I hope I will have a bit more time soon, thanks again for this PR, really sorry about my lack of time to try it…

torkelrogstad:
ping @fabien0102  :-)

micha-f:
@fabien0102 I added a note to the README 👍

PR #286, opened 2020-07-19
Support queryParamStringifyOptions on RestfulProvider level

### LINKED ISSUE #278: Support queryParamStringifyOptions on RestfulProvider level
**Is your feature request related to a problem? Please describe.**
I think the logic should be similar to `queryParams` (e.g. you can have `lang` which will be added to each request automatically). I'd like to have the same possibility for `queryParamStringifyOptions`, in my case I need to use `{ arrayFormat: 'brackets' }` for each request. It's not comfortable to specify this parameter each time I use array in `queryParams`.

**Describe the solution you'd like**
The same as we have with `queryParams`:
image

**Describe alternatives you've considered**
Currently, we can add `queryParamStringifyOptions` to each request where array is used. But as I mentioned, I don't like it :)

**Additional context**
I think it's clear enough already

p.s thanks for awesome library.

ApacheEx:
# Why

Resolves #278

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "restful-react",
-  "version": "14.2.2",
+  "version": "14.3.0",
   "description": "A consistent, declarative way of interacting with RESTful backends, featuring code-generation from Swagger and OpenAPI specs",
   "keywords": [
     "rest",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "restful-react",
-  "version": "14.2.1",
+  "version": "14.2.2",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "restful-react",
-  "version": "14.2.1",
+  "version": "14.2.2",
```
fabien0102, line 3: Since we add a new feature, don't be shy, you can bump the minor number 😉

ApacheEx: thx, I had the same feeling but wasn't sure) Fixed

Issue #289, opened 2020-08-06
Is there a way to enforce encoding for path parameters when using generated hooks?


brotheroftux:
Path parameters in the generated code are getting simply interpolated into the resulting path string, which is used in an actual request without any URI encoding (https://github.com/contiamo/restful-react/blob/454db37e029fdd3b8fceb98a6dc977720a91cb0b/src/scripts/import-open-api.ts#L301). Is there a way to somehow enforce encoding for path parameters without having to either write a custom generator or wrap every parameter value in `encodeURIComponent` at call site?

Thanks in advance.

fabien0102:
Seams reasonable to wrap everything in the generator, do you want to open a PR for this?

brotheroftux:
Sure, if by wrapping everything in generator you mean just adding `encodeURIComponent` to the built-in generator's code  for every path parameter. I'm not quite sure if this feature needs to be configurable &mdash; there might be a use case where a developer would want to provide, say, their own encoding function &mdash; so maybe we might want to add this option for the config.

fabien0102:
I think it's a reasonable behavior, I don't really see any drawback of using encodeURI everytime. 
If somebody have a need for more config over there, let's iterate, but for now I'm in favor of keep it simple 😁

brotheroftux:
Right, I'll open a PR later today then.

brotheroftux:
@fabien0102 We've got a case where we need the encoding to be RFC 3986 compliant (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent), so I guess I'll be adding configuration support for custom encode functions after all.

PR #290, opened 2020-08-07
Wrap path parameters with encodeURIComponent

brotheroftux:
# Why

See #289.

@fabien0102 Not certain if this really deserves a minor version bump. Also I reckon I should wrap the body for the DELETE mutation request here (https://github.com/contiamo/restful-react/blob/454db37e029fdd3b8fceb98a6dc977720a91cb0b/src/useMutate.tsx#L134), wasn't too sure about this though, will amend if needed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "restful-react",
-  "version": "14.4.0",
+  "version": "14.5.0",
   "description": "A consistent, declarative way of interacting with RESTful backends, featuring code-generation from Swagger and OpenAPI specs",
   "keywords": [
     "rest",
--- a/src/scripts/import-open-api.ts
+++ b/src/scripts/import-open-api.ts
@@ -474,11 +474,12 @@ ${description}export const ${componentName} = (${
       : `
     verb="${verb.toUpperCase()}"`
   }
-    path={\`${route}\`}${
+    path={encode\`${route}\`}${
     customPropsEntries.length
       ? "\n    " + customPropsEntries.map(([key, value]) => `${key}=${value}`).join("\n    ")
       : ""
   }
+    ${verb === "delete" ? "pathInlineBodyEncode={encodingFn}" : ""}
     {...props}
   />
 );
@@ -496,17 +497,19 @@ ${description}export const use${componentName} = (${
     verb === "get" ? "" : `"${verb.toUpperCase()}", `
   }${
     paramsInPath.length
-      ? `(paramsInPath: ${componentName}PathParams) => \`${route.replace(/\$\{/g, "${paramsInPath.")}\``
-      : `\`${route}\``
+      ? `(paramsInPath: ${componentName}PathParams) => encode\`${route.replace(/\$\{/g, "${paramsInPath.")}\``
+      : `encode\`${route}\``
   }, ${
-    customPropsEntries.length || paramsInPath.length
+    customPropsEntries.length || paramsInPath.length || verb === "delete"
       ? `{ ${
           customPropsEntries.length
             ? `${customPropsEntries
                 .map(([key, value]) => `${key}:${reactPropsValueToObjectValue(value || "")}`)
                 .join(", ")},`
             : ""
-        }${paramsInPath.length ? `pathParams: { ${paramsInPath.join(", ")} },` : ""} ...props }`
+        }${verb === "delete" ? "pathInlineBodyEncode: encodingFn, " : " "}${
+          paramsInPath.length ? `pathParams: { ${paramsInPath.join(", ")} },` : ""
+        } ...props }`
       : "props"
   });
 
@@ -536,7 +539,7 @@ export const Poll${componentName} = (${
       paramsInPath.length ? `{${paramsInPath.join(", ")}, ...props}` : "props"
     }: Poll${componentName}Props) => (
   <Poll<${genericsTypes}>
-    path={\`${route}\`}
+    path={encode\`${route}\`}
     {...props}
   />
 );
@@ -738,6 +741,24 @@ Path    : ${i.path}`),
 };
 
 /**
+ * Get the url encoding function to be aliased at the module scope.
+ * This function is used to encode the path parameters.
+ *
+ * @param mode Either "uricomponent" or "rfc". "rfc" mode also encodes
+ *             symbols from the `!'()*` range, while "uricomponent" leaves those as is.
+ */
+const getEncodingFunction = (mode: "uriComponent" | "rfc3986") => {
+  if (mode === "uriComponent") return "encodeURIComponent";
+
+  return `(uriComponent: string | number | boolean) => {
+  return encodeURIComponent(uriComponent).replace(
+      /[!'()*]/g,
+      (c: string) => \`%\${c.charCodeAt(0).toString(16)}\`,
+  );
+};`;
+};
+
+/**
  * Main entry of the generator. Generate restful-react component from openAPI.
  *
  * @param options.data raw data of the spec
@@ -753,6 +774,7 @@ const importOpenApi = async ({
   customImport,
   customProps,
   customGenerator,
+  pathParametersEncodingMode,
 }: {
   data: string;
   format: "yaml" | "json";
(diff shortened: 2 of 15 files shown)


brotheroftux:
Also I think it will be a nice touch to reduce the overall file size by aliasing `encodeURIComponent` for the module scope. I’ll do just that a bit later.

package.json
```diff
@@ -45,7 +45,7 @@
     "build": "run-p build:*",
     "build:project": "tsdx build",
     "build:bin": "tsc && rollup -c rollup.config.js",
-    "postbuild": "rimraf dist/**/*.test.*",
+    "prepare": "npm run build",
```
fabien0102, line 48: Please don't touch this, the deployment to npm is made by github actions 😉 

And this specific line is to avoid having the test part of the npm package (it's an ugly hack, but this is working ^^)

src/scripts/import-open-api.ts
```diff
@@ -737,6 +740,21 @@ Path    : ${i.path}`),
   // tslint:enable:no-console
 };
 
+/**
+ * Get the url encoding function to be aliased at the module scope.
+ * This function is used to encode the path parameters.
+ *
+ * @param mode Either "uricomponent" or "rfc". "rfc" mode also encodes
```
fabien0102, line 747: Do we really need this "rfc" mode? (you are the first one asking for this kind feature, so let's keep it simple)

fabien0102:
And sorry, you may have some git conflicts 🙄

package.json
brotheroftux: I think I deleted it by accident, sorry 😔

src/scripts/import-open-api.ts
brotheroftux: The thing is, our backend developer insists that we encode the parameters in accordance with RFC 3986. This means `!` and a bit more symbols have to be encoded, too. The biggest issue, however, is that I don't really see a clean, nice way to customize the encoding function. It all boils down to these options IMO:

- We can provide an `encodingFunction` option, and its value is a string, which is going to be interpolated into the resulting output file. This is the most effortless solution, however, this is very poor DX
- Same but the value of the option is a function. However we need to get the function's source code into the resulting file somehow, and we to get it typed. This seems not really feasible
- Provide an option `encodingFnAlias`, which when used in conjunction with `customImports` option will achieve the desired effect, i.e. you first declare a custom import with a custom encoding function in it, then specify to the code generator that you intend to use by using the `encodingFnAlias` option.
- Provide an option `encodingFn` with value of type `{ module: string; namedImport: string }`. This will ensure that a module containing a custom encoding function will get imported and used accordingly. However, this is a bit complicated.

So I decided to settle on two modes for now.

fabien0102: Fair enough! Let's stick with this, but let's maybe call the option `rfc3986` to avoid any missunderstanding :wink:

package.json
fabien0102: Happens, this is why we have PR reviews, don't worry 😃

fabien0102:
@brotheroftux You need to rebase against the `master` of this repository to resolve the conflicts 😉 

https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/syncing-a-fork

brotheroftux:
> @brotheroftux You need to rebase against the `master` of this repository to resolve the conflicts 😉
> 
> https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/syncing-a-fork

I know, I will do it just a bit later though. I still have to rename the mode anyway.

fabien0102:
> > @brotheroftux You need to rebase against the `master` of this repository to resolve the conflicts 😉
> > https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/syncing-a-fork
> 
> I know, I will do it just a bit later though. I still have to rename the mode anyway. (And it's getting a bit late here)

No problem, was just in case 😃 Have a good night

brotheroftux:
@fabien0102 I just realised that we need to use the same customized encoding function in `Mutate.tsx` and `useMutate.tsx`, which doesn't seem trivial.

In this case we can either a) add the same option to the `RestfulProvider` component and encode DELETE request's body as path parameter accordingly, or b) move the encoding logic to the fetcher components from the code generation.

Second solution implies that every library consumer, regardless of whether they're using the code generation utility or not, will get path parameter encoding. This, however, requires an obvious PR overhaul. Codewise this probably has to be implemented by splitting the path prop by `/` and encoding every path part with the `encodeURIComponent` or a user-provided function. This also implies that not only path parameters will get encoded, but everything between the slashes in a path string.

fabien0102:
mmmm, we could maybe move this logic to the `requestOptions`, we already have a `resolvePath` function with some options, I think adding an option in the provider should be easier, not trivial, but easier ^^

And this will follow the same pattern as the querystring option, so kind of consistant.

brotheroftux:
Yes, I agree. This also would be more consistent. At first I thought that maybe we only need this logic in the code generation utility, since library consumers who don't use it can actually utilize their own path generation functions &mdash; that's the reason why I decided to only encode the parameters in the generated code.

I am still however concerned that we would probably need to encode all path parts to make it work. Hopefully this won't cause any overencoding static parts of a path string.

Also I am not seeing `resolvePath` in type definitions for `RequestInit` and `RestfulReactProviderProps`, maybe I am looking in a wrong place.

fabien0102:
https://github.com/contiamo/restful-react/blob/5c3bf6684a1a5a589c7ea0975804edae299c8b8a/src/useGet.tsx#L73-L86

I was quickly looking, `useMutate` is using this function to resolve the path:

https://github.com/contiamo/restful-react/blob/5c3bf6684a1a5a589c7ea0975804edae299c8b8a/src/useMutate.tsx#L134-L139

With some tweak, this could be a good entry point for this kind of formatting, not totally sure yet, but this was an idea ^^

brotheroftux:
This is actually starting to get on my nerves. I just realised this fails when a path parameter contains an unencoded `/`.

fabien0102:
> @fabien0102 This is actually starting to get on my nerves. I just realised this fails when a path parameter contains an unencoded `/`.
> 
> I am out of ideas at this point. Maybe, just maybe, we should preprocess path args in the generator script by pre-encoding slashes somehow. No idea.

To be fair, I never had issue on the paths encoding, the default browser behaviours works well for me 🙄, so if you find a solution, even with not perfect dx (since this is for now what I call an edge case) I think this will be fine 😉 And if this doesn't fit, just fork, was nice to give a try, but sometimes... It actually make more sense to fork to put some custom logic.

Anyway, let's see, we may still have options to make this happen here 😀

brotheroftux:
Well, what about this version now?

fabien0102:
> Well, what about this version now?

If this is resolving your issue, this version looks good, if not, I guess we should move on to another approch… (It's sad, I really like your PR, but if this is not solving your issue, this is still increasing the complexity for "nothing" 😕 )

brotheroftux:
We'll test things a bit for our project, but as far as I am concerned, this does actually fix the issue. I changed things up once more, so now it should be fine.

fabien0102:
If everything follow, let's merge this, always nice to support more usecases :) 

Please note that you have some failing tests(snapshots, seams legit) 

image

brotheroftux:
@fabien0102 I updated the snapshots, so I think we can merge this.

package.json
fabien0102: @brotheroftux You still need to revert this change 

PS: Sorry for the latency of response, I'm in vacations ^^

brotheroftux: Removed the `prepare` script, `postbuild` was there. To be fair, I'd say we should let `prepare` stay, since it allows forks to be used as an NPM dependency (https://docs.npmjs.com/files/package.json#git-urls-as-dependencies).

fabien0102 (approved):
Looks good, thanks again for this amazing contribution! (I did resolved the conflicts myself ;) )

brotheroftux:
> Looks good, thanks again for this amazing contribution! (I did resolved the conflicts myself ;) )

Oh, there were conflicts? I checked yesterday I think &mdash; maybe I didn't notice.

Thanks!

fabien0102:
> > Looks good, thanks again for this amazing contribution! (I did resolved the conflicts myself ;) )
> 
> Oh, there were conflicts? I checked yesterday I think — maybe I didn't notice.
> 
> Thanks!

Was conflicting another PR that I merged just before, don't worry ;)

PR #294, opened 2020-08-23
Add params in function requestOptions

LeeChSien:
# Why

In some usecases, we would like add new values into request header which are depend on the request content.

For example, we tried to implement authentication method [oauth-dpop](https://tools.ietf.org/html/draft-ietf-oauth-dpop-01), but we found we don't have enough information in function requestOptions (request's url, method)

The original function:
```ts
() => Partial<RequestInit>
```

I modified this function like this:
```ts
(url: string, method: string, requestBody?: string) => Partial<RequestInit>)
```

So that we can achieve our purpose.
```ts
<RestfulProvider
  base="String!"
  resolve={data => data}
  requestOptions={async (url, method) => {
    // generate JWT with request's url and method.
    const dpopProofJWT = await DPoP(keypair, 'ES256', url, method);
    return { headers: { DPoP: dpopProofJWT } }
  }}
/>;
```

Please let me know if you have any concern, thank you!

### DIFF
--- a/src/useGet.test.tsx
+++ b/src/useGet.test.tsx
@@ -574,7 +574,9 @@ describe("useGet hook", () => {
     });
 
     it("should merge headers with providers", async () => {
-      nock("https://my-awesome-api.fake", { reqheaders: { foo: "bar", bar: "foo" } })
+      nock("https://my-awesome-api.fake", {
+        reqheaders: { foo: "bar", bar: "foo", visitUrl: "https://my-awesome-api.fake/", visitMethod: "GET" },
+      })
         .get("/")
         .reply(200, { oh: "my god 😍" });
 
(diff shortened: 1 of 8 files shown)


src/useGet.test.tsx
```diff
@@ -585,7 +587,12 @@ describe("useGet hook", () => {
       };
 
       const { getByTestId } = render(
-        <RestfulProvider base="https://my-awesome-api.fake" requestOptions={() => ({ headers: { bar: "foo" } })}>
+        <RestfulProvider
+          base="https://my-awesome-api.fake"
+          requestOptions={(url, method) => {
+            return { headers: { bar: "foo", visitUrl: url, visitMethod: method } };
```
LeeChSien, line 593: Add new test here.

LeeChSien:
I sent a PR to `restful-react`: https://github.com/contiamo/restful-react/pull/294
Add missing paramters in function requestOptions, in order to get url and method from request.

LeeChSien:
I sent a PR to `restful-react`: #294
Add missing paramters in function requestOptions, in order to get url and method from request.

fabien0102 (approved):
Looks good, thanks for your contribution! (I will publish a new version as soon as I review all other PRs 😉)

PR #295, opened 2020-08-24
add option to generate only typescript

bjornoj:
# Why

This PR will add an option to _only_ generate typescript types. I'm sure other libraries do this, but it's useful to have the functionality in this library as well.

For example it's useful for generating types for publishing npm packages.


fabien0102:
Sounds like a reasonable idea, thanks for this contribution.

One test is failing, let's fix this and merge this PR! 😃

src/scripts/import-open-api.ts
```diff
@@ -774,6 +915,7 @@ const importOpenApi = async ({
 
   resolveDiscriminator(specs);
 
+  let outputImports = `/* Generated by restful-react */\n\n${customImport ?? ""}`;
```
fabien0102, line 918: The previous version was a bit better with the indentation:

```tsx
${customImport ? `\n${customImport}\n` : ""}
```

- No extra `\n` if we don't have any `customImport`
- A trailing `\n` to avoid `customImport` to be merge with the other imports

src/scripts/import-open-api.ts
```diff
@@ -782,43 +924,54 @@ const importOpenApi = async ({
...
+          output += generateTypescriptApi(
+            verb,
+            operation.responses,
+            route,
+            operation.operationId,
+            operation.requestBody,
+            operation.parameters,
+          );
+        } else {
+          output += generateRestfulComponent(
```
fabien0102, line 937: This looks like of a lot of duplication over there, I will be more in favor of adding a param to `generateRestfulComponent` to skip the jsx (and this is maybe a goodtime to merge the param under an `options` object to avoid mistake)

src/scripts/import-open-api.ts
```diff
@@ -782,43 +924,54 @@ const importOpenApi = async ({
...
+          );
+        }
       }
     });
   });
 
-  const haveGet = Boolean(output.match(/<Get</));
-  const haveMutate = Boolean(output.match(/<Mutate</));
-  const havePoll = Boolean(output.match(/<Poll</));
+  if (!ts) {
```
fabien0102, line 952: This `if` should be useless, since we check than `<Get` is part of the generated code 😉

fabien0102 (changes requested):

fabien0102:
Managed with `--skip-react` flag introduce in #301

Issue #298, opened 2020-09-18
The github releases are not in sync with npm versions


cbek:
**Is your feature request related to a problem? Please describe.**
Currently, all pushes on master branch ends in npm publish.
This leads to inconsistency between [npm packages ](https://www.npmjs.com/package/restful-react/)and [GitHub](https://github.com/contiamo/restful-react/releases).

**Describe the solution you'd like**
Only git tags leads to an npm publish.

fabien0102:
The solution wanted for this issue:

> The flow I would like to have in place is the following:
>
> * any contributor can bump the version on the package.json in a PR
> * as soon as this PR is merged, if the package.json was updated -> publish to npm
> * draft a github release with a nice changelog **<- the missing part!**

https://github.com/contiamo/restful-react/pull/299#issuecomment-703469233

fabien0102:
As we can see here -> https://github.com/contiamo/restful-react/releases/tag/v15.0.0

This is done! Now every publish to npm will be followed by a github release (with the associated changelog) 🎉

cbek:
Good work. ❤️

PR #299, opened 2020-09-18
#298 Version releases without git tags

cbek:
Only git tags leads to an npm publish.

# Why
https://github.com/contiamo/restful-react/issues/298

### DIFF
--- a/.github/workflows/npm-publish.yaml
+++ b/.github/workflows/npm-publish.yaml
@@ -2,8 +2,8 @@ name: Npm publish
 
 on:
   push:
-    branches:
-      - master
+    tags:
+      - '*'
     paths:
       - package.json
 


fabien0102:
Hello @cbek thanks for the contribution, I do get the problem and it's a known issue (due to my previous lazyness 🙄), I will try to add a github realease step in the process and this should solve the issue 😃 

The flow I would like to have in place is the following:
 - any contributor can bump the version on the `package.json` in a PR
 - as soon as this PR is merged, if the `package.json` was updated -> publish to npm
 - draft a github release with a nice changelog **<- the missing part!**

So goal is to an have easy npm publish flow, without the need to do it manually, the drawback is that we loose some git tags in the process, but this could be resolve with a github release.

PR #301, opened 2020-09-28
Skip react generation

fabien0102:
# Why

Add a `--skip-react` flag, this will skip all react component generation but keep the custom generator part.
This will replace the #295 approch and permit more usecase (as generating code for a nodejs or angular setup).

## Todo
- [x] Deal with unused `encode` function generation
- [x] Add integration test


fabien0102:
image

Little goodies feature, let's get back some type safety in the advanced config 😃

mpotomin (approved):
Looks good 👍

PR #302, opened 2020-10-04
Adding the ability to not fetch on a useGet

shockdesign:
# Why

Adds the ability to skip, handy when missing a token or using this for authentication and don't want it to run without that data.

### DIFF
--- a/src/useGet.test.tsx
+++ b/src/useGet.test.tsx
@@ -506,6 +506,27 @@ describe("useGet hook", () => {
     });
   });
 
+  describe("with skip", () => {
+    it("should not fetch", async () => {
+      const children = jest.fn();
+      children.mockReturnValue(<div />);
+
+      const MyAwesomeComponent = () => {
+        const { data, error, loading } = useGet<{ oh: string }>({ path: "/", skip: true });
+        return children({ data, error, loading });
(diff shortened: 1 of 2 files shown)


fabien0102:
Hello @shockdesign, thanks for your contribution, but normally the `lazy` props have the exact same behaviour in this `skip` 😉 (But don't worry, this is a commun mistake / question 😁)

shockdesign:
@fabien0102 Is that the case, I found lazy just to delay the call until after the view was rendered. This can omit the call even after render to be called on a later update when appropriate changes are updated on that view.

For example I have an auth endpoint which requires the username and password to be passed via the header. However we want to wait until those conditions are met.

fabien0102:
Yes, `lazy` behave exactly like this (since you are familiar with the code, you can check the details ^^).

The `lazy` props was designed to prevent any calls before calling `fetch`, but in practice you can control this props as you will do with a `skip` and this is doing the same thing 😉

shockdesign:
And looking at it, @fabien0102 it looks like you're correct. `lazy` just seems inappropriately named as you need actually enforce `lazy` as false in order for your fetch to occur.

However it works, so closing this PR.

PR #303, opened 2020-10-07
Clear the `data` after an error

fabien0102:
# Why

Having some stale data after error is really confusing and causing some issues.
We also don't have any good use cases in mind where having the previous data can be useful, so let's simplify the library behaviour.


mpotomin (dismissed):
Makes sense to me 👌

Issue #297, opened 2020-08-25
Response headers.. Again..


singulart:
When working with useGet, an access to response headers is highly desirable, as usually backends send pagination-related info not in JSON body but in response headers.

**Describe the solution you'd like**
Ideally, access to response headers should be available through `response` object
```
  const {data: getPetsApiResponse, response, loading, error} = useGetPets(
    {
      queryParams:  { page: page}
    }
  );
```

**Describe alternatives you've considered**
I've also tried the global onResponse as per https://github.com/contiamo/restful-react/pull/254, but unfortunately, response headers are empty there, as well.

abhinavrastogi-harness:
@singulart I have opened a PR for returning response headers as I also need access to headers for individual requests.

However, for the global `onResponse`, are you trying to read `response.headers` directly? `response.headers` implements the Headers class, which is an iterable, and not a plain object.

See https://developer.mozilla.org/en-US/docs/Web/API/Headers.

You should be able read the headers if you write `response.headers.forEach` or  use `for..of`. 
Individual headers can be retrieved via `response.headers.get()`

PR #307, opened 2020-10-08
Add support for returning original response

abhinavrastogi:
# Why

This will allow the user to pass the flag `orginalResponse` to `useGet` to enable returning the original response object.
This solves the requirement where a user needs to access the response headers per request. 
Putting this behind a new flag keeps the API backward-compatible and opt-in for performance reasons (since the response object can be pretty large).

Related issues: 
- https://github.com/contiamo/restful-react/issues/297
- https://github.com/contiamo/restful-react/issues/173

This is different from `onResponse` added in https://github.com/contiamo/restful-react/pull/254 as this allows per-request handling of response headers instead of global handling.

I have already added tests to showcase how this works. I can add this to the Readme also once this gets reviewed.


abhinavrastogi:
I opened same PR from my work account by mistake (#306). Closed that one and opened this one now. Apologies for the spam.

src/useGet.test.tsx
```diff
@@ -645,6 +645,63 @@ describe("useGet hook", () => {
...
+  describe("originalResponse", () => {
+    it("should return original response when flag is enabled", async () => {
+      nock("https://my-awesome-api.fake")
+        .get("/")
+        .reply(200, { oh: "my god 😍" });
+
+      const MyAwesomeComponent = () => {
+        const { loading, response } = useGet({
+          path: "/",
+          originalResponse: true,
```
fabien0102, line 657: Why this flag? I would expect to always send the response

src/useGet.test.tsx
```diff
@@ -645,6 +645,63 @@ describe("useGet hook", () => {
...
+      };
+
+      const { getByTestId } = render(
+        <RestfulProvider base="https://my-awesome-api.fake">
+          <MyAwesomeComponent />
+        </RestfulProvider>,
+      );
+
+      await waitForElement(() => getByTestId("response"));
+      expect(getByTestId("response")).not.toBeEmpty();
```
fabien0102, line 674: Could be nice to add some header or something to make sure everything is working as expected

fabien0102 (changes requested):
I will suggest to remove the flag, this is adding complexity for no benefit (at least I don't see any right now)

Also, can you add the same feature for `<Get />`? This should be quite straight forward, so could be nice to maintain some compatibility for people than can't use hooks

src/useGet.test.tsx
abhinavrastogi: I assumed that returning response was left out on purpose since it's part of the TS interface, but not actually returned anywhere. If that's not the case, I can definitely remove the flag.

I am a bit apprehensive about always returning the entire response by default because it can have performance and memory-usage implication. Returning the response means that the data is duplicated (once in parsed form and once as a Blob in response).

Another option I considered was to just return the response headers separately (and continue returning null for response). That will solve a common request for headers but it'll not be a very clean API.

What do you think?

fabien0102: I'm not too afraid about memory over there, will be garbage collect quite quickly if you don't consume it and we can still add the flag later if we really have some memory issues.

I actually like the `responseHeaders` option, especially if we can bring back some types over there 😀 But I'm not against adding the `response` for advanced usages, even if this can't really be typesafe. This is always good to have the information around.

abhinavrastogi: Cool. I'll remove the flag then. It'll be much cleaner then.

abhinavrastogi:
@fabien0102 can you please review again and guide me on the next steps? I have made all the requested changes.

fabien0102 (commented):
Looks good! 💯 Nice job, just a little `npm version minor` on this PR and we are good to go (so you will have a release as soon as we merge the PR on master 😃)

abhinavrastogi:
@fabien0102 running `npm version minor` is rewriting the entire changelog with reference to my fork. I assume that's not right.

fabien0102:
> @fabien0102 running `npm version minor` is rewriting the entire changelog with reference to my fork. I assume that's not right.

Indeed, I forgot about this issue… Don't worry, I will take care of this 😉 (and try to find something to solve this issue ^^)

fabien0102:
I didn't have the permission to push on your repo so I did a follow up branch -> #310

fabien0102:
@abhinavrastogi Thanks again for you contribution! This will be merge soon 👍

PR #311, opened 2020-10-14
Fix `path` generation

fabien0102:
# Why

I did introduced a bug in https://github.com/contiamo/restful-react/pull/301/commits/617d4f604b2bfa5cffb82c8733603c125be6d559#diff-2e6f87c49da8750d7ff365e15e082df94de0f5a63b5b414fa6f3c8908420e27dR359

The `path` need to be a template string to works as expected.

I also fixed the `Poll` component that had the `encode` function even without custom encoder.


src/scripts/tests/__snapshots__/import-open-api.test.ts.snap
```diff
@@ -495,7 +495,7 @@ export type UpdatePetProps = Omit<MutateProps<Pet, Error, void, UpdatePetRequest
 export const UpdatePet = ({id, ...props}: UpdatePetProps) => (
   <Mutate<Pet, Error, void, UpdatePetRequestRequestBody, UpdatePetPathParams>
     verb=\\"PATCH\\"
-    path=\\"/pets/\${id}\\"
+    path={\`/pets/\${id}\`}
```
fabien0102, line 498: This was the bug

mpotomin (approved):

PR #318, opened 2020-11-06
fix: Avoid including undefined bodies in DELETE

torkelrogstad:
There's a bug, where calling `useDelete(...).mutate()` with no body would add the string "undefined" to the path we're querying. We fix this by checking if the body is undefined before converting it.


fabien0102 (approved):
Looks good, sorry I did missed this notification 🙄
I will publish a new version with this inside, thanks for this fix 😃

PR #329, opened 2021-02-01
Include spec version in generated file.

micha-f:
# Why

Including the version of the OpenAPI / Swagger spec that was used to generate fetchers to support with API Versioning.


fabien0102 (approved):
Looks good

PR #335, opened 2021-03-18
Stabilize the identity of the mutate function

amacleay-cohere:
Hopefully begins to address #186

Approach:
* Remove (mutable) references to state. Instead, hook into the
  (immutable) `setState` function in the mutate function body where
  state references are necessary
* Try very hard to be honest about what dependencies may change in the
  dependency array, but leave only immutable dependencies that depend only
  on props (almost totally successful)

# Why

Hopefully addresses some of #186 , at least with respect to the mutate function.

### DIFF
--- a/src/useMutate.test.tsx
+++ b/src/useMutate.test.tsx
@@ -418,6 +418,37 @@ describe("useMutate", () => {
     });
   });
 
+  describe("Mutate identity", () => {
+    it("should remain the same across calls", async () => {
+      nock("https://my-awesome-api.fake")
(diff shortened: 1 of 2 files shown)


kfujita-cohere (approved):

kfujita-cohere (changes requested):
LGTM but accidentally selected approve when going to comment so I'm changing to **Request changes** until you see this

fabien0102 (approved):
Just amazing! Thank you!

PR #337, opened 2021-03-23
Revert "Stabilize the identity of the mutate function"

amacleay-cohere:
Reverts contiamo/restful-react#335

I realized while testing this with a much larger project today that this PR adds a problem: if the identity of the mutate function _does_ happen to change, the `abort()` function no longer gets called with the prior instance's abort handler, but the current instance's abort handler! I am working on a fix for this, but in the meantime, #335 should just be reverted in a bugfix release.

### DIFF
--- a/src/useMutate.test.tsx
+++ b/src/useMutate.test.tsx
@@ -418,37 +418,6 @@ describe("useMutate", () => {
     });
   });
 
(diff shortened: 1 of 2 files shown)


amacleay-cohere:
prefer #338

PR #338, opened 2021-03-23
Stabilize use mutate even more

amacleay-cohere:
Revisiting #335 in light of the issue noted in #337

Stabilize the useMutate function even more by using a deep
compare effect. This better handles cases in which the argument to
`useMutate` is an inline object, which is the typical use case.

### DIFF
--- a/src/useMutate.test.tsx
+++ b/src/useMutate.test.tsx
@@ -490,8 +490,8 @@ describe("useMutate", () => {
       await result.current.mutate({});
       const mutate2 = result.current.mutate;
 
-      expect(mutate0).toEqual(mutate1);
-      expect(mutate0).toEqual(mutate2);
+      expect(mutate0).toBe(mutate1);
+      expect(mutate0).toBe(mutate2);
     });
   });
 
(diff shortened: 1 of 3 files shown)


src/useMutate.test.tsx
```diff
@@ -447,6 +465,34 @@ describe("useMutate", () => {
...
+          wrapper,
+        },
+      );
+      const mutate0 = result.current.mutate;
+      const mutate1 = result.current.mutate;
+      await result.current.mutate({});
+      const mutate2 = result.current.mutate;
+
+      expect(mutate0).toEqual(mutate1);
+      expect(mutate0).toEqual(mutate2);
```
fabien0102, line 494: Is it working with `.toBe` instead of `.toEqual`? (`.toEqual` don't insure that the ref is the same https://jestjs.io/docs/expect#toequalvalue)

amacleay-cohere: Does indeed! Pushed that fix.

fabien0102 (approved):
👌

PR #346, opened 2021-04-15
Support form-data on mutate function (already fixed for useMutate)

### LINKED ISSUE #264: Generated API add text/plain to content-type header request
**Describe the bug**
Generated API hook is adding `text/plain` to `content-type` header. 

**To Reproduce**
Generate API using the next definition:
``` yaml
post:
    tags:
        - security
        - legacy
    responses:
        '200':
            content:
                application/json:
                    schema:
                        $ref: '#/components/schemas/MenuItems'
            description: Success
    operationId: postGetMenu
```

Headers are set as follows:
``` ts
  const myHook = usePostGetMenu({
    requestOptions: { headers: { 'Content-Type': 'application/json', Accept: 'application/json' } }
  })
```
And I make the request:
``` ts
    myHook.mutate().then((value: MenuItems) => {
      console.log(value)
    })
```

**Expected behavior**
I expect the request header to have `content-type: application/json` and not `content-type: text/plain, application/json`
image

**Desktop (please complete the following information):**
 - OS: Windows 10 Enterprise
 - Browser Chrome 80

Thanks in advance.

dcruzin:
This fix is a continuation of #257 and #168. 

It applies the same changes made on `useMutate` to the `mutate` function.

Without this fix, `Content-Type` is still being filled as `text/plain` when it should be set as `multipart/form-data` by the browser.

This also fixes #264, by not trying to set the default Content-Type if it's already set on `mutateRequestOptions.headers`.

### DIFF
--- a/src/Mutate.tsx
+++ b/src/Mutate.tsx
@@ -193,19 +193,24 @@ class ContextlessMutate<TData, TError, TQueryParams, TRequestBody, TPathParams> 
 
     const request = new Request(makeRequestPath(), {
       method: verb,
-      body: typeof body === "object" ? JSON.stringify(body) : body,
+      body: body instanceof FormData ? body : typeof body === "object" ? JSON.stringify(body) : body,
       ...(typeof providerRequestOptions === "function"
         ? await providerRequestOptions<TRequestBody>(makeRequestPath(), verb, body)
         : providerRequestOptions),
       ...mutateRequestOptions,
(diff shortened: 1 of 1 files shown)


fabien0102 (approved):

fabien0102:
@dcruzin Thanks for this PR, should arrive soon in npm (`v15.4.2`)

PR #348, opened 2021-04-17
stabilize the identity of useGet and its refetch method

### LINKED ISSUE #186: Lazy fetching of data will cause infinite loop / runaway effect

amacleay-cohere:
Stabilizing the identity of the `fetchData` function presents some
challenges. The biggest is that `fetchData` needs to have access to
`state` and `setState`, but it needs to have the most up-to-date version
of them in order to work. There are two ways to have done this: the
first is to move the definition of `fetchData` into the function body of
`useGet` itself and make it a closure over the `state` and `setState`
variables - that is the approach taken here. Alternatively, we could
define `fetchData` as a function which takes `state` and `setState` as
parameters, but is closed over a particular combination of props and
context values - this option wasn't explored.

The end result is that the identity of the `refetch` function is stable
given a certain set of props and context, and it is stable over a deep
comparison thereof - so both of the following work fine:
```typescript
const queryParams = { page: 1 };
function MyComponent = () => {
  const { refetch } = useGet({ queryParams, lazy: true })`
  ...
};
```

```typescript
function MyComponent = () => {
  const { refetch } = useGet({ { page: 1 }, lazy: true })`
  ...
};
```

This should resolve #186 fully (I hope!)

# Why
Solves #186

## NB

There are 3 tests which suffer from the same symptom: for a reason I haven't completely tracked down, there is an extra re-render dispatched by the hook with identical state: in both instances, it is when the re-render is called, there are 2 returns out of the hook with `loading: true`. I don't suppose this will cause any issues with any callers, but I was unable to totally root it out.

Note that this uses the same `useDeepCompareCallback` that is also present in #338 - the implementation in either case is identical and could be rebased out safely depending on merge order.

### DIFF
--- a/src/useGet.test.tsx
+++ b/src/useGet.test.tsx
@@ -421,9 +421,16 @@ describe("useGet hook", () => {
 
       await wait(() => expect(children).toBeCalledTimes(2));
 
+      expect(children.mock.calls[0][0].loading).toEqual(true);
+      expect(children.mock.calls[0][0].error).toEqual(null);
+
+      expect(children.mock.calls[1][0].loading).toEqual(false);
+      expect(children.mock.calls[1][0].error).toMatchObject({ status: 404 });
+
       fireEvent.click(getByTestId("set-page-button"));
 
       await wait(() => expect(children).toBeCalledTimes(5));
+
       expect(children.mock.calls[1][0].error).not.toEqual(null);
       expect(children.mock.calls[4][0].loading).toEqual(false);
       expect(children.mock.calls[4][0].data).toEqual({ id: 1 });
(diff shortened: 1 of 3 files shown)


fabien0102:
Sorry, I missed the notification, will review this tomorrow!

Just as a sidenote, I did saw this on twitter recently: https://epicreact.dev/the-latest-ref-pattern-in-react/ and of course, I thought about this old issue ^^ This could be neat way to solve this!

src/useGet.test.tsx
```diff
@@ -719,6 +726,48 @@ describe("useGet hook", () => {
...
+      expect(children.mock.calls[1][0].data).toEqual(null);
+
+      // finally resolved
+      expect(children.mock.calls[2][0].loading).toEqual(false);
+      expect(children.mock.calls[2][0].data).toEqual({ id: 1 });
+
+      // If the refetch identity weren't stable, this would loop off forever and ever...
+      // you'll know you broke this test if your computer fan goes crazy!
+      expect(children.mock.calls[0][0].refetch).toEqual(children.mock.calls[1][0].refetch);
+      expect(children.mock.calls[0][0].refetch).toEqual(children.mock.calls[2][0].refetch);
```
fabien0102, line 768: Same as your other PR, I would love to try with `toBe` instead of `toEqual` for this usecase

src/useGet.test.tsx
```diff
@@ -403,12 +403,12 @@ describe("useGet hook", () => {
 
       const MyAwesomeComponent: React.FC<{ path: string }> = ({ path }) => {
         const [page, setPage] = useState(1);
-        const params = useGet<{ id: number }, any, { page: number }>({ path, queryParams: { page } });
+        const useGetReturn = useGet<{ id: number }, any, { page: number }>({ path, queryParams: { page } });
```
fabien0102, line 406: `use` prefix indicate a hook, please find a better name

src/useGet.test.tsx
```diff
@@ -945,9 +994,10 @@ describe("useGet hook", () => {
         </RestfulProvider>,
       );
 
-      await wait(() => expect(children).toBeCalledTimes(3));
-      expect(children.mock.calls[2][0].loading).toEqual(false);
-      expect(children.mock.calls[2][0].data).toEqual({ id: 1 });
+      // A mystery... this is called an extra time
```
fabien0102, line 997: Even if this not a big issue, a bit of investigation over there could be nice (at least to understand ^^)

src/useGet.test.tsx
amacleay-cohere: Fixed! https://github.com/contiamo/restful-react/compare/d8998875e04df253c51d9024acb96dd86837cbf6..3f02d9da77cff55d425e1fd05ede3652b4f559b4

src/useGet.test.tsx
amacleay-cohere: Fixed! https://github.com/contiamo/restful-react/compare/d8998875e04df253c51d9024acb96dd86837cbf6..3f02d9da77cff55d425e1fd05ede3652b4f559b4

fabien0102 (approved):
Looks good, let's ship this!
Thanks again for this amazing contribution! 🎉

Issue #359, opened 2021-06-02
Stable Code Generation.


vkbansal:
**Is your feature request related to a problem? Please describe.**
Our swagger.json is generated dynamically and hence the order of objects is not guaranteed.
This creates a large diff every time because the interfaces and their keys are always in random order.

**Describe the solution you'd like**
I'd like all the definitions to be sorted by names. (even the keys of the interfaces and string enums)

fabien0102:
Hi @vkbansal,
I'm sorry, but, from the library perspective, the code generation is stable (same input -> same output).
I don't know how your `swagger.json` is generated, but you open an issue on this side, I don't really any quick win to have an ordered input here.

(this said, if you have an idea of API & implementation, we will always be happy to collaborate on a pull-request)

vkbansal:
I have an idea on how to implement it. I was wondering if you will be open to such a feature if I open a PR. I do not want to start working on it, if it is not going to be accepted.

fabien0102:
If this is an option in `restful-react.config.js` and don't break the "natural" order, I'm totally fine with this 👍

PR #360, opened 2021-06-02
feat: sort interfaces by name

vkbansal:
# Why

solves #359

### DIFF
--- a/src/bin/restful-react-import.ts
+++ b/src/bin/restful-react-import.ts
@@ -23,6 +23,7 @@ export interface Options {
   transformer?: string;
   validation?: boolean;
   skipReact?: boolean;
+  sorted?: boolean;
 }
 
 export type AdvancedOptions = Options & {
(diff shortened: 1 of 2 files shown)


vkbansal:
@fabien0102 This is how it will work.

Also I would like to do the same here https://github.com/contiamo/restful-react/blob/master/src/scripts/import-open-api.ts#L151

fabien0102:
> @fabien0102 This is how it will work.
> 
> Also I would like to do the same here https://github.com/contiamo/restful-react/blob/master/src/scripts/import-open-api.ts#L151
> 
> Can you suggest a way to do it?

This is indeed quite deep, will be a pain to add an option here… 😕 Otherwise, another approch will be to add a postprocess step, taking the generated file as input, and reordering using typescript compiler directly (https://github.com/Microsoft/TypeScript/wiki/Using-the-Compiler-API#re-printing-sections-of-a-typescript-file)

fabien0102:
https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/member-ordering.md

Or, you can maybe use an eslint rule (with `--fix`) on this file

vkbansal:
I guess I can also use `transform` options to pre-process the json using [fast-json-stable-stringify](https://www.npmjs.com/package/fast-json-stable-stringify)

fabien0102:
> I guess I can also use `transform` options to pre-process the json using [fast-json-stable-stringify](https://www.npmjs.com/package/fast-json-stable-stringify)

Indeed! Smart idea!

PR #362, opened 2021-06-30
Add support for anyOf

x539:
OpenAPI schemas with `anyOf[A,B]` should generate a type `A | B` not `{}`.

https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/


abonifacio (approved):

fabien0102 (changes requested):
Good catch, just a little request to be able to add this `anyOf` support, please add some unit tests (so we don't break this in the future without noticing it)

Thanks for this great addition!

fabien0102:
Merged into #370 , I did rebased and add some unit test, thanks for your contribution!

PR #363, opened 2021-07-10
Add support for tuples

abonifacio:
# Why

Currently if you generate types based on the following definition, `age_range` is typed as `number[]`

```yaml
age_range:
  description: 'A range of passenger ages (start and end, inclusively)'
  type: array
  items:
    type: number
  minItems: 2
  maxItems: 2
```

As the array should only contain two items, it would be nice if it was typed as `[number, number]`

## Note

This will conflict with #362, but if you support the idea, I'll rebase afterwards


fabien0102 (approved):
Nice idea! I love it, thank for your contribution 💯

PR #364, opened 2021-07-13
Support required at higher level for allOf and oneOf

abonifacio:
# Why

Given the following spec

```yaml
components:
  schemas:
    User:
      type: object
      properties:
        name:
          type: string
        email:
          type: string
    RegisteredUser:
      type: object
      allOf:
        - $ref: "#/components/schemas/User"
      required:
        - name
        - email
```

Resulting types are

```ts
interface User {
  name?: string;
  email?: string;
}
type RegisteredUser = User;
```

With this change, resulting types are 

```ts
type Require<T,R extends keyof T> = T & Required<Pick<T, R>>;
interface User {
  name?: string;
  email?: string;
}
type RegisteredUser = Require<User, "name" | "email">;
```

This adds a helper type (`Require<T,R extends keyof T>`) to make it more readable, otherwise the output would be

```ts
type RegisteredUser = User & Required<Pick<User,  "name" | "email">>;
```


fabien0102 (approved):
Looks good, thanks for this great addition! 👌

PR #367, opened 2021-07-15
Keep last path parameter when skipping React

### LINKED ISSUE #366: Final path parameter in DELETE requests is excluded from generated interfaces in `--skipReact` mode
**Describe the bug**

The final path parameter is excluded from the generated `*PathParams` interface

**To Reproduce**
1. Create a path with a path parameter in the last segment of the path
2. Define a GET and a DELETE operation for the path, properly defining all path parameters
3. Generate a `.ts` file using `restful-react --skipReact`
4. Observe the parameter's presence in the GET request, but not in the DELETE

An example reproduction can be found [in this repository](https://github.com/s-thom/restful-react-366-reproduction).

**Expected behavior**

I expect all path parameters to be defined in the generated interfaces.

**Additional context**

This is related to #334, which also found this case, but determined that it was a feature. In the cases of generating components, [as described in the comment](https://github.com/contiamo/restful-react/blob/688e0729623e38b7ff936d0dcedb89998c85088f/src/scripts/import-open-api.ts#L332-L347) I tend to agree with this idea (though documentation would be nice). However, in the case of `--skipReact`, we won't have the `<Delete />` component passing a function to fill in this last parameter. I believe it is more intuitive to include all parameters in the PathParams interface. 

I suggest that this change only applies to `--skipReact` mode, so that it does not affect the primary use-case of this library. It would be a breaking change, as there may be others using `--skipReact` that are relying on this behaviour.

s-thom:
Resolves #366 (see issue for motivation).

Adds all path parameters to generated interfaces for DELETE requests when `--skipReact` is set.

### DIFF
--- a/src/scripts/import-open-api.ts
+++ b/src/scripts/import-open-api.ts
@@ -344,7 +344,9 @@ export const generateRestfulComponent = (
    *    </DeleteResource>
    */
 
-  const paramsInPath = getParamsInPath(route).filter(param => !(verb === "delete" && param === lastParamInTheRoute));
+  const paramsInPath = getParamsInPath(route).filter(
+    param => !(!skipReact && verb === "delete" && param === lastParamInTheRoute),
+  );
   const { query: queryParams = [], path: pathParams = [], header: headerParams = [] } = groupBy(
     [...parameters, ...(operation.parameters || [])].map<ParameterObject>(p => {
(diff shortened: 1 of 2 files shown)


fabien0102 (approved):
👌 Very nice, works as described!

PR #368, opened 2021-07-15
support enum on type number

abonifacio:
# Why

Given the following spec

```yaml
description: A representation of a floor of a bus.
title: Floor
type: object
properties:
  level:
    description: The floor number.
    enum:
      - 1
      - 2
    type: number
```

I would like the type to be generated as 

```ts
interface Floor {
  level: 1 | 2;
}
```

Additional question: I don't use React but this is the best library I found to generate type from OpenAPI specs, is there a plan to split this library into type generation & react related stuff?

### DIFF
--- a/src/scripts/import-open-api.ts
+++ b/src/scripts/import-open-api.ts
@@ -52,7 +52,7 @@ export const getScalar = (item: SchemaObject) => {
     case "long":
     case "float":
     case "double":
-      return "number" + nullable;
+      return (item.enum ? `${item.enum.join(` | `)}` : "number") + nullable;
 
     case "boolean":
       return "boolean" + nullable;
--- a/src/scripts/tests/import-open-api.test.ts
+++ b/src/scripts/tests/import-open-api.test.ts
@@ -66,6 +66,7 @@ describe("scripts/import-open-api", () => {
       { item: { type: "int64" }, expected: "number" },
       { item: { type: "float" }, expected: "number" },
       { item: { type: "number" }, expected: "number" },
+      { item: { type: "number", enum: [1, 2] }, expected: `1 | 2` },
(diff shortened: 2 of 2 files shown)


micha-f:
@abonifacio We use this library in many places without react (e.g. in many nodejs services). Have you seen the `skipReact` option in the config? https://github.com/contiamo/restful-react#only-generating-custom-code-no-react-hookscomponents . If you don't specify a custom generator, it will only output types.

abonifacio:
@micha-f yeah totally! I'm using `skipReact` and everything works beautifully. I was just thinking that maybe if you split the library

1. It will be simpler for non React users (less dependencies, less documentation, less intimidating to contribute)
1. You can give it a more catchy name, currently the name doesn't imply that you can generate types from OpenAPI!

micha-f (approved):

micha-f:
You are right, the name does not communicate very well that this project can do quite a bit more these days. It started out as a project to make interacting with REST APIs from react more comfortable. Then @fabien0102 got to work and added all of this awesome OpenAPI stuff 🤷 . As it powers almost every service / UI we put into production it should be pretty well-tested and capable by now. @fabien0102 certainly had the ambition to make this the best library to generate types from specs 🙂 . 

Not sure we can change the project structure without much effort at this point though. Let's see what @fabien0102 thinks.

In the meantime - this PR looks good 👍 .

fabien0102 (approved):
This PR looks good indeed, thanks for this nice addition! 🤘 

Regarding splitting the react and OpenAPI generation part, I'm thinking of this since quite a long time. We could build a very nice generation framework indeed (I may already have some POC locally 😇). But time… Maybe one day! 😁

Issue #268, opened 2020-05-24
refetch should return proper value in lazy get


smilexu:
**Is your feature request related to a problem? Please describe.**
I`d like to use react-select async select component with restful-react. For async loading options, we need to return proper options via loadOptions callback, the code is like below.

```
<AsyncSelect
  ...
  loadOptions={(input, callback) => {
    const result  = doQueryByInput(input);
    callback(buildOptionsFromResult(result));
  }}
/>
```
The problem is that refetch will not return any value, I have no idea how to return proper options by callback.

**Describe the solution you'd like**
I'd like to get value from refetch, code like below

```
  const { data, refetch } = useGet({
    path: '/search-something',
    lazy: true,
  });

...

<AsyncSelect
  ...
  loadOptions={(input, callback) => {
    refetch({
      queryParams: buildQueryParams(input),
    }).then((response) => {
      callback(buildOptions(response));
    });
  }}
/>
```

**Describe alternatives you've considered**
No other alternatives so far, but if you have any idea can archive that, it would be also appreciate.

**Additional context**
No

fabien0102:
Hello,
we are also using react-select internally, so I just checked 😋 

We are using a simple fetch with a custom generator instead of a react hook.

Something like this:
https://github.com/contiamo/restful-react/blob/61604ee50626b3c910224225cf4dd92cf4b260b0/examples/restful-react.config.js#L20-L40

But this is mainly to have types from the open-api specs, otherwise a simple fetcher is more appropriate than restful-react for this usecase ;) (you just need to export your `base` to compose your url)

smilexu:
Not sure if I understood your comment correctly, I already implement my project based on restful-react which handles error, auth etc, do not want to add a more solution (use other fetcher)only for async load scenario. So I will prefer to have some restful-react build-in solution to solve that if possible. `apollo-graphql` is a good example, we can reset params when refetching, but sure, it is different implementation.

fabien0102:
Make sense, for now `refetch` already take some arguments, but is totally bind to the state (because it's a refetch, not a fetch).

This would be easy to return some value from `refetch` (I just checked the codebase), but will this make the API a bit funky, since refetch was meant to refetching, not fetching… (so you will have some useless rerender for the `loading` for example)

I sadly don't have a lot of time right now to play with this, but you can try to play around locally and submit a PR. I think a cleaner solution will be to introduce a `useFetch`, to avoid the conflict between the two API (data from state vs data from promise).

Maybe something like this
[...]

PR #376, opened 2021-09-29
Allow fetchData in useGet to return data (so refetch can return data)

amacleay-cohere:
# Why

Attempts to solve #268 by following the non-suggested path: muddle the `useGet` API by allowing `refetch` to return the data it just fetched, rather than making the caller put their code in an ambient `useEffect`

```tsx
# Uncomfortable
const { data, refetch } = useGet("/mycollection")
useEffect(() => {
  if (data.results.length > 1) {
    alert("There are several");
  }
}, [data]);
return (
  <button onClick={async () => {
    await refetch();
  }} />
);
```

```tsx
# Much clearer and closer to the `mutate` API
const { data: originalData, refetch } = useGet("/mycollection")

return (
  <button onClick={async () => {
    if ((await refetch()).results.length > 1) {
      alert("Several");
    }
  }} />
)
```

### DIFF
--- a/src/useGet.tsx
+++ b/src/useGet.tsx
@@ -99,7 +99,7 @@ export interface UseGetReturn<TData, TError, TQueryParams = {}, TPathParams = un
   /**
    * Refetch
    */
-  refetch: (options?: RefetchOptions<TData, TError, TQueryParams, TPathParams>) => Promise<void>;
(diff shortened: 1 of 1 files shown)


fabien0102 (approved):
This is looking nice 👌

Issue #333, opened 2021-02-24
React v17 support


kazuki:
I used a restful-react generated code with React v17 and works fine.
But latest npm raises error `ERESOLVE unable to resolve dependency tree` without `--legacy-peer-deps`.

Could you add react v17 version range in peerDependencies ?
`$ yarn test` passes all tests with React v17.

singulart:
+1 on this

sabzDeveloper:
+1

jdmeyer3:
+1

PR #378, opened 2021-11-04
Initial support for React 17

disrupted:
# Why

I added React 17 to peer dependencies as per #333 to allow updating.


fabien0102 (approved):

fabien0102:
@micha-f Can you have a look, I don’t have access to netlify 😇

Issue #380, opened 2021-12-20
Fix allOf in responseBodies generation


dcruzin:
**Describe the bug**

Same issue as #317 and #319, but for `responseBodies`

PR #381, opened 2021-12-20
Fix allOf  in responseBodies generation

dcruzin:
# Why

#380


fabien0102 (approved):
Looks good to me, @micha-f can you have a look also?

micha-f:
@dcruzin @fabien0102 released in https://github.com/contiamo/restful-react/releases/tag/v15.9.4

